PR #1, opened 2020-04-16
parse SVG gradient coordinates and transform for COLRv1

anthrotype:
I will add more tests and comments tomorrow, but you can take a look already if you like.


anthrotype:
this current PR requires the other one from https://github.com/rsheeter/nanosvg/pull/2 to be merged and released first

PR #11, opened 2020-04-30
Svg colr svg

rsheeter:
Rough in an initial svg > colr > svg test. Will be much more interesting once we have colrv1 and gradients.


tests/colr_to_svg.py
```diff
@@ -23,6 +23,12 @@
 from typing import Tuple
 from fontTools.pens import svgPathPen
 from fontTools.pens import transformPen
+from fontTools.ttLib.tables import otTables
+
+
+_PAINT_SOLID = 1
```
anthrotype, line 29: we could even make this an IntEnum inside nanoemoji.paint module and use if for the various Paint*.Format class attributes

tests/colr_to_svg.py
```diff
@@ -38,60 +44,112 @@ def _svg_root(view_box: Rect) -> etree.Element:
     return svg_root
 
 
-def _colr_v0_to_svg(
+def _nice_path(svg_pen: svgPathPen.SVGPathPen) -> str:
+    # svgPathPen doesn't format #s nicely
```
anthrotype, line 48: maybe we can change svgPathPen to our liking

tests/colr_to_svg.py
```diff
@@ -0,0 +1,155 @@
...
+
+_PAINT_SOLID = 1
+_PAINT_LINEAR_GRADIENT = 2
+_PAINT_RADIAL_GRADIENT = 3
+
+
+def _emsquare_to_viewbox(upem: int, view_box: Rect):
+    if view_box != Rect(0, 0, view_box.w, view_box.w):
+        raise ValueError("We simply must have a BOX from 0,0")
+    return color_glyph.map_viewbox_to_emsquare(Rect(0, 0, upem, upem), view_box.w)
```
anthrotype, line 37: the viewbox might not be square.

how about :

```
Affine2D.rect_to_rect(Rect(0, 0, upem, upem), Rect(view_box.x, view_box.y + view_box.h, view_box.w, -view_box.h)
```

(untested)

anthrotype: this is even better `Affine2D.rect_to_rect(Rect(0, upem, upem, -upem), view_box)`

anthrotype:
nevermind my earlier comment about non-square viewbox. Of course it's square, we're converting from font units to svg units, and the font's em-square is, well.. square.

tests/colr_to_svg.py
rsheeter: Works for me but beware that this means formatting svg the way I like it for little reason other than I like it that way :D

PR #45, opened 2020-06-09
Add shape reuse for glyf and svg

### LINKED ISSUE #44: Implement shape reuse for svg
For `glyf` we can identify and take advantage of in and inter glyph shape recurrence. Explore doing the same for svg. Ref https://docs.microsoft.com/en-us/typography/opentype/spec/svg#glyph-identifiers.

Clock faces (https://rsheeter.github.io/android_fonts/emoji.html?q=note:o%E2%80%99clock) and shaded hands (https://rsheeter.github.io/android_fonts/emoji.html?q=u:270d) should make decent test cases.

rsheeter:
Comparing sizes is no fun if some teams get a leg up.

Fixes #44


src/nanoemoji/svg.py
```diff
@@ -0,0 +1,189 @@
...
+def _add_glyph(svg, color_glyph, id_updates):
+    # each glyph gets a group of its very own
+    svg_g = svg.append_to("/svg:svg", etree.Element("g"))
+    svg_g.attrib["id"] = color_glyph.glyph_name
+
+    # copy the shapes into our svg
+    glyphs = {}
+    for painted_layer in color_glyph.as_painted_layers():
+        reuse_key = _inter_glyph_reuse_key(painted_layer)
+        if reuse_key not in glyphs:
```
anthrotype, line 107: You are resetting the `glyphs` dict every time you call `_add_glyph`, instead of initializing it in the caller frame, and passing it down and accummulating in `_add_glyph`.
So this is not actually sharing/reusing layers with same path/paint _between_ color_glyphs, only _within_ each color_glyph -- am I correct?
If so, the use of a function named `_inter_glyph_reuse_key` (instead of _intra_) is kind of misleading..

But should we not also share layers between glyphs that belong to the same disjoint set? I thought that was the point of grouping them in `_glyph_groups`, but I might be missing something very basic.. 🤔

src/nanoemoji/svg.py
```diff
@@ -0,0 +1,189 @@
...
+    SVG reuses w/paint so paint is part of key."""
+
+    # TODO we could recycle shapes that differ only in paint, would just need to
+    # transfer the paint attributes onto the use element if they differ
+    return (painted_layer.paint, painted_layer.path.d, painted_layer.reuses)
+
+
+def _add_glyph(svg, color_glyph, id_updates):
+    # each glyph gets a group of its very own
+    svg_g = svg.append_to("/svg:svg", etree.Element("g"))
```
anthrotype, line 100: perhaps we could do away with a "g" wrapper if it's the only glyph in the svg document

src/nanoemoji/color_glyph.py
```diff
@@ -219,7 +219,7 @@ class ColorGlyph(NamedTuple):
     ufo: ufoLib2.Font
     filename: str
     glyph_name: str
-    glyph_id: str
+    glyph_id: int
```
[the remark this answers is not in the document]
rsheeter: I _think_ it's because our type annotations are not consistently present so there are holes in the analysis

src/nanoemoji/disjoint_set.py
```diff
@@ -0,0 +1,53 @@
...
+    # union by rank
+    def union(self, x: T, y: T):
+        x_root = self.find(x)
+        y_root = self.find(y)
+        if x_root == y_root:
+            return  # already in the same set
+        if self.rank[x] < self.rank[y]:
+            tmp = x_root
+            x_root = y_root
+            y_root = tmp
```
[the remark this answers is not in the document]
rsheeter: Oh Python, you are so wonderful.

src/nanoemoji/svg.py
```diff
@@ -0,0 +1,189 @@
...
+        curr_id: str = gradient.attrib["id"]
+        new_id = f"{color_glyph.glyph_name}::{curr_id}"
+        del gradient.attrib["id"]
+        gradient_xml: str = etree.tostring(gradient)
+        if gradient_xml in id_updates:
+            id_updates[curr_id] = id_updates[gradient_xml]
+        else:
+            gradient.attrib["id"] = new_id
+            id_updates[curr_id] = new_id
+            id_updates[gradient_xml] = new_id
```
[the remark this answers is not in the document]
rsheeter: I had two and that was kind of ugly so I made it one. I think perhaps the name is just wrong.

src/nanoemoji/svg.py
```diff
@@ -0,0 +1,189 @@
...
+from nanoemoji.color_glyph import ColorGlyph, PaintedLayer
+from nanoemoji.disjoint_set import DisjointSet
+from picosvg.svg import to_element, SVG
+from picosvg import svg_meta
+from picosvg.svg_transform import Affine2D
+import regex
+from typing import MutableMapping, Sequence, Tuple
+
+
+def _create_doclist(svg: SVG) -> str:
```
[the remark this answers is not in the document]
rsheeter: Deleted!

src/nanoemoji/svg.py
rsheeter: Ugh no that's a bug I caused when teasing apart an overlong function. For bonus points I have a test that specifically looks for this and "corrected" it's contents w/o noticing this had broken.

rsheeter: Fixed. It's unfortunately fairly easy to miss what's changing in the test ttx because the svg content is not pretty printed.

src/nanoemoji/svg.py
```diff
@@ -0,0 +1,189 @@
...
+        # establish base svg, defs
+        svg = SVG.fromstring(
+            r'<svg version="1.1" xmlns="http://www.w3.org/2000/svg"><defs/></svg>'
+        )
+        svg_defs = svg.xpath_one("//svg:defs")
+        for color_glyph in (color_glyphs[g] for g in group):
+            _add_unique_gradients(id_updates, svg_defs, color_glyph)
+            _add_glyph(svg, color_glyph, id_updates)
+
+        # print(etree.tostring(svg.svg_root, pretty_print=True).decode("utf-8"))
```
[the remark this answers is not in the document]
rsheeter: Done

src/nanoemoji/svg.py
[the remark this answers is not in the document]
rsheeter: Split them and then grouped all the reusable parts into a single class. Better I think.

src/nanoemoji/svg.py
rsheeter: I kind of like the consistency. It makes predicting what our compiler does very easy. You have to have the id somewhere so this is just costing you the actual `<g></g>`.

src/nanoemoji/svg.py
```diff
@@ -23,7 +23,17 @@
...
+from typing import MutableMapping, NamedTuple, Sequence, Tuple
+
+
+class ReuseCache(NamedTuple):
+    old_to_new_id: MutableMapping[str, str]
+    gradient_to_id: MutableMapping[str, str]
+    shapes: MutableMapping[str, etree.Element]
+
+    @classmethod
+    def create(cls):
```
anthrotype, line 35: the fact that this contains _mutable_ dicts makes this an awkward tuple. How about we turn it into a dataclass, and do `old_to_new_id: Dict[str, str] = dataclasses.field(default_factory=dict)`, etc. to address the mutability issue of the default `{}` values?

src/nanoemoji/nanoemoji.py
```diff
@@ -270,16 +272,30 @@ def _draw_glyph_extents(ufo: ufoLib2.Font, glyph: Glyph):
 
 
 def _glyf_ufo(ufo, color_glyphs):
```
anthrotype, line 274: while working on #52 I came across the fact that "glyf" as a `--color_format` and `_glyf_ufo` as function name in here don't make much sense when --output_format is .otf (i.e. CFF-flavored OpenType). The name `glyf` is tied to TrueType. We want to convey the idea that this is _not_ a layered font (like COLR is), and that it is _not_ colored as it only has black & white (or if you like, only black). How about nouns like "flat", or "black"? Or even just "simple" or "normal"..

src/nanoemoji/nanoemoji.py
```diff
@@ -270,16 +272,30 @@ def _draw_glyph_extents(ufo: ufoLib2.Font, glyph: Glyph):
...
+    # glyphs by reuse_key
+    glyphs = {}
+    swaps = []
```
anthrotype, line 277: this `swaps` local variable seems unused

src/nanoemoji/nanoemoji.py
```diff
@@ -270,16 +272,30 @@ def _draw_glyph_extents(ufo: ufoLib2.Font, glyph: Glyph):
...
+            # if we've seen this shape before reuse it
+            reuse_key = _inter_glyph_reuse_key(painted_layer)
+            if reuse_key not in glyphs:
+                glyph = _create_glyph(color_glyph, painted_layer)
+                glyphs[reuse_key] = glyph
+            else:
+                glyph = glyphs[reuse_key]
+            parent_glyph.components.append(Component(baseGlyph=glyph.name))
+
+        # No great reason to keep single-component glyphs around
```
anthrotype, line 296: it's that single-component glyphs are wasteful, but rather the fact that a given glyph is only ever referenced as a component in only one other composite glyph (and is not itself a primary glyph with a direct cmap or GSUB mapping, but it's only reason of existence is that it is used as component).
E.g. if glyph C is referenced by both A and B, even if the latter two only have one component each, you're still reusing C in both, thus saving a bit of space. But if glyph C (and D, E, F, etc.) is (are) only used as component(s) in _one_ other glyph, say, A, then you are wasting space to keep them as individual glyphs and A may better be a simple, decomposed glyph.

src/nanoemoji/nanoemoji.py
```diff
@@ -270,16 +272,30 @@ def _draw_glyph_extents(ufo: ufoLib2.Font, glyph: Glyph):
...
             ufo.info.familyName,
             color_glyph.glyph_name,
-            svg_units_to_font_units,
+            color_glyph.transform_for_font_space(),
         )
+        parent_glyph = ufo.get(color_glyph.glyph_name)
         for painted_layer in color_glyph.as_painted_layers():
-            _create_glyph(color_glyph, painted_layer)
+            # if we've seen this shape before reuse it
+            reuse_key = _inter_glyph_reuse_key(painted_layer)
```
anthrotype, line 288: in theory, at least for glyf (probably not for COLR), these inter-glyph reuses need not be exactly the same `path.d` but could be normalized first in a similar way as we do for in-glyph reuses, and then referenced as components with a translation offset. I guess we could do that later.

src/nanoemoji/nanoemoji.py
```diff
@@ -270,16 +272,30 @@ def _draw_glyph_extents(ufo: ufoLib2.Font, glyph: Glyph):
...
+            if reuse_key not in glyphs:
+                glyph = _create_glyph(color_glyph, painted_layer)
+                glyphs[reuse_key] = glyph
+            else:
+                glyph = glyphs[reuse_key]
+            parent_glyph.components.append(Component(baseGlyph=glyph.name))
+
+        # No great reason to keep single-component glyphs around
+        if len(parent_glyph.components) == 1:
+            ufo[color_glyph.glyph_name] = ufo[parent_glyph.components[0].baseGlyph]
```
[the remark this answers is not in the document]
rsheeter: Fixed.

src/nanoemoji/svg.py
rsheeter: I think my theory was I didn't want the actual dict to be replaced. In looking at it again +1 to dataclasses. I left the type as MutableMapping.

src/nanoemoji/nanoemoji.py
rsheeter: ha I think I thought I'd swap something. Fixed!

src/nanoemoji/nanoemoji.py
rsheeter: I think that may only make sense to add if we add transform to COLR layers. glyf can't be a color format so I don't want to optimize it all that much.

src/nanoemoji/nanoemoji.py
rsheeter: Migrating to a separate issue.

Issue #30, opened 2020-05-28
nuke 'post' glyph names


rsheeter:
As per `# TODO may wish to nuke 'post' glyph names`

PR #48, opened 2020-06-09
Drop postscript glyph names for TTFs; add --keep-glyph-names option

anthrotype:
Fixes https://github.com/googlefonts/nanoemoji/issues/30

[ufo2ft 2.15](https://github.com/googlefonts/ufo2ft/releases/tag/v2.15.0) added support for dropping postscript glyph names from TTFs by setting the compiled font's `post` table to format 3.0.

For CFF-flavored OTF, ufo2ft still doesn't allow to remove glyph names, as these are embedded in the CFF table itself and required. Maybe in the future we will be able to convert from name-keyed CFF to CID-keyed CFF (haven't looked into it yet).

This PR also adds a `--nokeep_glyph_names` flag to the nanoemoji script. I decided not to drop glyph names by default as this would change all the TTX test files, besides it's good to keep the glyph names for debugging purposes and only drop them when making final fonts.

### DIFF
--- a/src/nanoemoji/nanoemoji.py
+++ b/src/nanoemoji/nanoemoji.py
@@ -59,6 +59,7 @@ class ColorFontConfig(NamedTuple):
     family: str
     color_format: str
     output_format: str
+    keep_glyph_names: bool = True
 
 
 class InputGlyph(NamedTuple):
@@ -111,6 +112,9 @@ flags.DEFINE_string(
     "/tmp/AnEmojiFamily-Regular.ttf",
     "Dest file, can be .ttf, .otf, or .ufo",
 )
+flags.DEFINE_bool(
+    "keep_glyph_names", True, "Whether or not to store glyph names in the font."
+)
 
 
 def _codepoints_from_filename(filename):
@@ -137,7 +141,7 @@ def _inputs(filenames: Iterable[str]) -> Generator[InputGlyph, None, None]:
             yield InputGlyph(filename, codepoints, picosvg)
 
 
-def _ufo(family: str, upem: int) -> ufoLib2.Font:
+def _ufo(family: str, upem: int, keep_glyph_names: bool = True) -> ufoLib2.Font:
     ufo = ufoLib2.Font()
     ufo.info.familyName = family
     # set various font metadata; see the full list of fontinfo attributes at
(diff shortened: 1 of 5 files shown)


src/nanoemoji/nanoemoji.py
```diff
@@ -111,6 +112,9 @@ class ColorGenerator(NamedTuple):
     "/tmp/AnEmojiFamily-Regular.ttf",
     "Dest file, can be .ttf, .otf, or .ufo",
 )
+flags.DEFINE_bool(
+    "keep_glyph_names", True, "Whether or not to store glyph names in the font."
```
rsheeter, line 116: Suggest False

src/nanoemoji/nanoemoji.py
```diff
@@ -151,6 +155,10 @@ def _ufo(family: str, upem: int) -> ufoLib2.Font:
     space.width = upem
     ufo.glyphOrder = [".notdef", ".space"]
 
+    if not keep_glyph_names:
+        # use 'post' format 3.0 for TTFs, shaving a kew KBs of unneeded glyph names
+        ufo.lib[ufo2ft.constants.KEEP_GLYPH_NAMES] = keep_glyph_names
```
rsheeter, line 160: why do we need an if? - just always set = keep_glyph_names

tests/make_emoji_font_test.py
```diff
@@ -73,3 +73,23 @@ def test_make_emoji_font(svgs, expected_ttx, color_format, output_format):
     )
     _, ttfont = nanoemoji._generate_color_font(config, glyph_inputs)
     test_helper.assert_expected_ttx(svgs, ttfont, expected_ttx)
+
+
+@pytest.mark.parametrize(
+    "svgs", [("rect.svg", "rect2.svg"), ("one-o-clock.svg", "two-o-clock.svg")]
```
rsheeter, line 79: is this what ./precommit.sh gave? (runs black)

anthrotype: yes

src/nanoemoji/nanoemoji.py
anthrotype: yeah, that works too. I'll remove the if

PR #51, opened 2020-06-11
add support for CFF2

### LINKED ISSUE #49: support compiling OTF with CFF2 table
ufo2ft 2.15 supports compiling static (non-variable) OTF fonts with CFF2 table. The `compileOTF` function has a `cffVersion` parameter defaulting to 1, which one can now optionally set to 2.
The new ufo2ft also allows to subroutinize CFF2 fonts (via Adobe's [cffsubr](https://github.com/adobe-type-tools/cffsubr) tool).

CFF2 is more compact than CFF 1.0 as it removes some redundant data which is defined elsewhere in the OTF (widths, glyph names, etc.).
Subroutinization has a similar effect as our current shape-reuse approach for TTFs via composite glyphs. Potentially it could even yield better results, since the sharing can involve parts of charstrings, not just whole components like in TrueType. It would be interesting to compare file size of subroutinized CFF2-flavored COLR fonts with the TrueType variants that use composites.

One problem with implementing this in nanoemoji is that we guess the output format by checking the output file's extension (.ttf, .otf, or .ufo). But CFF1 or CFF2 fonts have the same file extension: `.otf`.

So I think we need an extra flag `--cff_version 2`.

anthrotype:
Fixes #49


anthrotype:
I'm not sure what kind of tests we'd like to have. Maybe this is enough to start playing with it and see if worth pursuing this route.

PR #54, opened 2020-06-12
Adjust flags

### LINKED ISSUE #53: Some combinations of flags don't make sense
As @anthrotype noted `--color_format` and `--output_file` have begun to overlap. For example, setting `--color_format glyf --output_file my.otf` feels odd. Clean this up.

> while working on #52 I came across the fact that "glyf" as a --color_format and _glyf_ufo as function name in here don't make much sense when --output_format is .otf (i.e. CFF-flavored OpenType). The name glyf is tied to TrueType. We want to convey the idea that this is not a layered font (like COLR is), and that it is not colored as it only has black & white (or if you like, only black). How about nouns like "flat", or "black"? Or even just "simple" or "normal"..

rsheeter:
Fixes #53

### DIFF
--- a/src/nanoemoji/nanoemoji.py
+++ b/src/nanoemoji/nanoemoji.py
@@ -80,19 +80,38 @@ class InputGlyph(NamedTuple):
 class ColorGenerator(NamedTuple):
     apply_ufo: Callable[[ufoLib2.Font, Sequence[ColorGlyph]], None]
     apply_ttfont: Callable[[ufoLib2.Font, Sequence[ColorGlyph], ttLib.TTFont], None]
+    font_ext: str  # extension for font binary, .ttf or .otf
 
 
 _COLOR_FORMAT_GENERATORS = {
-    "glyf": ColorGenerator(lambda *args: _glyf_ufo(*args), lambda *_: None),
-    "colr_0": ColorGenerator(lambda *args: _colr_ufo(0, *args), lambda *_: None),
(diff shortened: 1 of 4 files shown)


src/nanoemoji/nanoemoji.py
```diff
@@ -80,19 +80,38 @@ class InputGlyph(NamedTuple):
...
-    "svg": ColorGenerator(lambda *_: None, lambda *args: _svg_ttfont(*args, zip=False)),
-    "svgz": ColorGenerator(lambda *_: None, lambda *args: _svg_ttfont(*args, zip=True)),
+    "glyf": ColorGenerator(lambda *args: _glyf_ufo(*args), lambda *_: None, ".ttf"),
+    "glyf_colr_0": ColorGenerator(
+        lambda *args: _colr_ufo(0, *args), lambda *_: None, ".ttf"
+    ),
+    "glyf_colr_1": ColorGenerator(
+        lambda *args: _colr_ufo(1, *args), lambda *_: None, ".ttf"
+    ),
+    "cff_colr_0": ColorGenerator(
```
anthrotype, line 94: if we are adding the prefix `cff` to the --color_format, maybe we can get rid of the extra `--cff_version` option, and select the right CFF table version based on the --color_format prefix `cff_colr_1` vs `cff2_colr_1`

rsheeter: Done

PR #57, opened 2020-06-14
Make sure svg content is appropriately mapped into font space

### LINKED ISSUE #56: shape reuse lost viewBox
Shape reuse changes dropped viewbox off svgs, put it back

rsheeter:
Fixes #56


src/nanoemoji/nanoemoji.py
```diff
@@ -83,11 +83,33 @@ class ColorGenerator(NamedTuple):
 
 
 _COLOR_FORMAT_GENERATORS = {
+<<<<<<< Updated upstream
```
anthrotype, line 86: You forgot to remove these lines from stash merge conflict

src/nanoemoji/nanoemoji.py
```diff
@@ -83,11 +83,33 @@ class ColorGenerator(NamedTuple):
...
     "glyf": ColorGenerator(lambda *args: _glyf_ufo(*args), lambda *_: None),
     "colr_0": ColorGenerator(lambda *args: _colr_ufo(0, *args), lambda *_: None),
     "colr_1": ColorGenerator(lambda *args: _colr_ufo(1, *args), lambda *_: None),
     "svg": ColorGenerator(lambda *_: None, lambda *args: _svg_ttfont(*args, zip=False)),
     "svgz": ColorGenerator(lambda *_: None, lambda *args: _svg_ttfont(*args, zip=True)),
+=======
```
anthrotype, line 92: And here

src/nanoemoji/nanoemoji.py
```diff
@@ -83,11 +83,33 @@ class ColorGenerator(NamedTuple):
...
+    "cff_colr_1": ColorGenerator(
+        lambda *args: _colr_ufo(1, *args), lambda *_: None, ".otf"
+    ),
+    "svg": ColorGenerator(
+        lambda *_: None, lambda *args: _svg_ttfont(*args, compressed=False), ".ttf"
+    ),
+    "svgz": ColorGenerator(
+        lambda *_: None, lambda *args: _svg_ttfont(*args, compressed=True), ".ttf"
+    ),
+>>>>>>> Stashed changes
```
anthrotype, line 112: And here :)

rsheeter: Heh I thought I'd have time to fix that before you looked!

rsheeter:
Fixes #58 
Renders #57 irrelevant

anthrotype:
> Renders #57 irrelevant

you mean, "renders #56 irrelevant"?

anthrotype:
the PR title should be modified, since it is now deliberately not keeping the viewBox

src/nanoemoji/svg.py
```diff
@@ -83,24 +86,36 @@ def _add_unique_gradients(
             svg_defs.append(gradient)
 
 
-def _inter_glyph_reuse_key(painted_layer: PaintedLayer):
+# https://docs.microsoft.com/en-us/typography/opentype/spec/svg#coordinate-systems-and-glyph-metrics
+def _svg_matrix(transform: Affine2D) -> str:
+    # TODO handle rotation: units of Affine2D and svg matrix don't match
```
anthrotype, line 91: I'm confused by this TODO. Why do you mean by "handle rotation: units of Affine2D and svg matrix don't match" here?

anthrotype:
It took me a while to fully understand what's going on here. It gets immediately confusing once transformation matrices are involved.
Basically, IIUC, in an svg document as embedded in OT-SVG table the y axis is vertically flipped (positive y axis points downwards); so far so good, that's what we already knew from the svg spec and we account for that when we map from viewbox to emsquare units in the svg to COLR pipeline. However, the OT-SVG spec also says that y=0 is considered to be "the horizontal baseline for text layout". And further down, it says that (unless viewBox or translation matrix are specified) a glyph in OT-SVG is expected to be drawn in the +x -y quadrant of the SVG coordinate system. Our viewbox-to-emsquare transform has the effect of "moving" the svg (0, 0) origin to the point (0, UPEM) in font space; we assume that the horizontal baseline of the SVG glyph is positioned at viewBox.y+viewBox.h (in svg coordinate system), and that the SVG glyph is drawn in the +x +y positive SVG quadrant, in upright position.
OT-SVG wants the glyph to be drawn in +x -y quadrant and the baseline at y=0, so we need again to flip y axis and subtract the UPEM from y-coordinates (i.e. negating the `d` and `f` element of the viewbox-to-emsquare matrix).

I wonder if, instead of modifying the result of `transform_for_font_space()` in `_svg_matrix` function (also the name here is not very telling), we should have a new method that starts with viewBox and UPEM and returns a transform suitable for inclusion in OT-SVG table glyphs. One that would only apply the x/y scale and subtract the UPEM, e.g.

```python
def map_viewbox_to_OT_SVG_space(view_box: Rect, upem) -> Affine2D:
    # scale to font upem
    x_scale = upem / view_box.w
    y_scale = upem / view_box.h
    # shift so origin is 0,0
    dx = -view_box.x * x_scale
    dy = -view_box.y * y_scale
    # shift elements drawn in the +x +y quadrant of the SVG coordinate system so that
    # they appear in the +x -y quadrant, above the y=0 baseline of OT-SVG glyphs
    dy -= upem
    return Affine2D(x_scale, 0, 0, y_scale, dx, dy)
```

src/nanoemoji/svg.py
rsheeter: Updated to explicitly note radians vs degrees

src/nanoemoji/svg.py
```diff
@@ -88,8 +88,7 @@ def _add_unique_gradients(
 
 # https://docs.microsoft.com/en-us/typography/opentype/spec/svg#coordinate-systems-and-glyph-metrics
 def _svg_matrix(transform: Affine2D) -> str:
-    # TODO handle rotation: units of Affine2D and svg matrix don't match
-    transform = transform._replace(d=-transform.d, f=-transform.f)
+    # TODO handle rotation: Affine2D uses radiens, svg is in degrees
```
anthrotype, line 91:
```suggestion
    # TODO handle rotation: Affine2D uses radians, svg is in degrees
```

anthrotype: I still don't understand what this has to do with "rotation". We are only mapping two rectangles (the svg viewbox and the UPEM square), we are not rotating anything, nor dealing with angles

rsheeter: A correct implementation would work even if we had rotate. Agreed that it shouldn't actually be happening.

rsheeter: Work or fail; currently it would silently do the wrong thing.

anthrotype: I'm still confused, sorry. You are just stringifying the Affine2D matrix as an SVG `matrix(a b c d e f)` attribute. The latter comprises just 6 floats, there are no rotation angles involved, no degrees nor radians (the spelling "radiens" is still incorrect, but it doesn't matter). If you were to decompose the 6-tuple matrix in the concatenation of its components (scale, translate, skew/rotation) then rotation itself (and the metric of rotation angles) could become an issue. But I don't see it happening here, and to me the TODO doesn't add information, only leaves me wondering. But I'll stop now.

PR #65, opened 2020-06-19
Fun with bad data

rsheeter:
Noto Emoji have some funky stuff. Try to report it a bit more nicely when encountered.


rsheeter:
Merging with TODO per IM conversation

PR #71, opened 2020-06-27
Raw svg

rsheeter:
Add ability to produce a font with almost unmodified svgs


src/nanoemoji/nanoemoji.py
```diff
@@ -102,11 +102,17 @@ class ColorGenerator(NamedTuple):
...
-        lambda *_: None, lambda *args: _svg_ttfont(*args, compressed=False), ".ttf"
+    "picosvg": ColorGenerator(
+        lambda *_: None, lambda *args: _svg_ttfont(*args, picosvg=True, compressed=False), ".ttf"
     ),
-    "svgz": ColorGenerator(
-        lambda *_: None, lambda *args: _svg_ttfont(*args, compressed=True), ".ttf"
+    "picosvgz": ColorGenerator(
+        lambda *_: None, lambda *args: _svg_ttfont(*args, picosvg=True, compressed=True), ".ttf"
+    ),
+    "untouchedsvg": ColorGenerator(
```
anthrotype, line 111: why not just call it `"svg"`? The "untouched" seems unnecessary

PR #77, opened 2020-07-08
Fix radial gradient with transform 0 scale and non-0 skew

### LINKED ISSUE #76: radial gradient with 0 scale and non-0 skew disappear
in https://github.com/googlefonts/noto-emoji/blob/master/svg/emoji_u1f46b.svg, the brown hair of the lady on the left hand side are filled with a radial gradient that has this `gradientTransform="matrix(9.9137e-14 1 -.9288 9.9137e-14 39.072 -120.91)"`.
When xscale and yscale are zero (in this case they almost are) and the skewX and skewY are non-zero, the trasform is like swapping x with y and y with x (in the simple case [0 1 -1 0] or [0 -1 1 0] this are 90 degrees rotations).

Now, when converting to COLRv1 paint, the Affine2x2 produced by nanoemoji is a zero matrix (0,0,0,0), so we lose the skewX/Y part and the radial gradient is invalid.

That's because of this:
https://github.com/googlefonts/nanoemoji/blob/0e622ae42ff0e5071da0c620553f262a06b530a5/src/nanoemoji/color_glyph.py#L174-L176

We need to consider the `b` and `c` values of the Affine2D when deciding on the radius scale. The fix is taking max() of all the absolute a,b,c,d values, not just a and d.

anthrotype:
Fixes #76

### DIFF
--- a/src/nanoemoji/color_glyph.py
+++ b/src/nanoemoji/color_glyph.py
@@ -164,14 +164,14 @@ def _parse_radial_gradient(grad_el, shape_bbox, view_box, upem):
     c1 = transform.map_point(c1)
 
     # As for the circle radii (which are affected by Affine2x2), we only scale them
-    # by the largest between the horizontal and vertical (absolute) scale.
+    # by the maximum of the (absolute) scale or skew.
     # Then in Affine2x2 we only store a "fraction" of the original transform, i.e.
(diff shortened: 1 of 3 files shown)


anthrotype:
I'd like to merge this today, as I will be off next week. Feel free to fix up or open issues which I can attend to when I'm back the following week.

PR #78, opened 2020-07-08
flip y_scale sign before translating in map_viewbox_to_font_emsqure

anthrotype:
Otherwise we translate in the wrong direction.
This finally fixes the odd viewBox of the baseball noto-emoji (emoji_u26BE.svg), which as you remember doesn't start at 0,0.
I can see it properly after I roundtrip it back from COLRv1 to SVG (in progress #11).

### DIFF
--- a/src/nanoemoji/color_glyph.py
+++ b/src/nanoemoji/color_glyph.py
@@ -49,10 +49,10 @@ def _shift_origin_0_0(
 
 def map_viewbox_to_font_emsquare(view_box: Rect, upem: int) -> Affine2D:
     x_scale, y_scale = _scale_viewbox_to_emsquare(view_box, upem)
-    dx, dy = _shift_origin_0_0(view_box, x_scale, y_scale)
-
-    # flip y axis and shift so things are in the right place
+    # flip y axis
     y_scale = -y_scale
+    # shift so things are in the right place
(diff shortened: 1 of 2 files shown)


anthrotype:
I'm 100% sure this is the correct fix so I'll merge for now, we can reopen this issue later in case.
To convince oneself, see how using [`Affine2D.rect_to_rect`](https://github.com/googlefonts/picosvg/blob/a556ac8f8133e8acae31e38aaa4a8e40fd410bdc/src/picosvg/svg_transform.py#L148-L161) method gives the same results as doing each invidual computation in `map_viewbox_to_font_emsquare`, using the incriminated example of emoji_u26BE.svg's viewBox:

```
>>> map_viewbox_to_font_emsquare(Rect(-151, 297, 128, 128), 1024)
Affine2D(a=8.0, b=0, c=0, d=-8.0, e=1208.0, f=3400.0)
>>> Affine2D.rect_to_rect(Rect(-151, 297, 128, 128), Rect(0, 1024, 1024, -1024))
Affine2D(a=8.0, b=0, c=0, d=-8.0, e=1208.0, f=3400.0)
```

We are basically making so that the SVG origin (-151, 297) becomes the point (0, 1024) in UPEM space, with the y scale flipped (hence the -1024 sign) in the second rect.

PR #90, opened 2020-07-28
Use ninja

rsheeter:


src/nanoemoji/nanoemoji.py
```diff
@@ -489,6 +475,56 @@ def _generate_color_font(config: ColorFontConfig, inputs: Iterable[InputGlyph]):
...
+def resolve_rel_build(path):
+    return os.path.abspath(os.path.join(build_dir(), path))
+
+
+def write_preamble(nw):
+    nw.comment('Generated by nanoemoji')
+    nw.newline()
+
+    nw.rule('picosvg', 'picosvg $in > $out || echo "$in failed picosvg"')
+    nw.rule('codepointmap', 'nanoemoji_write_codepoints $in > $out')
```
anthrotype, line 505: how about instead of installing setuptools CLI scripts for write_codepoints and write_fea (just to make the executable by ninja rules), we invoke the modules as scripts with `python -m nanoemoji.write_codepoints`? That basically has the effect of importing a module with its name set to "__main__" (which will then trigger the main() entry point).

Actually, we want `f"{sys.executable} -m nanoemoji.write_codepoints"` rather than unqualified `python`, so that we will reuse the same exact interpreter that was used to run nanoemoji in the first place.

ninja/write_ninja.py
```diff
@@ -0,0 +1,65 @@
+# Experimental ninja
+
+from ninja import ninja_syntax
```
anthrotype, line 3: I presume you forgot to check in the repo the ninja_syntax.py module from https://github.com/ninja-build/ninja/blob/master/misc/ninja_syntax.py

There's also a `ninja_syntax` module that one can pip install from PyPI but I don't know if it's in sync with upstream or maintained (last updated in 2017).

rsheeter: ninja_syntax comes along with ninja.

src/nanoemoji/nanoemoji.py
rsheeter: Oh good idea, will update. I don't think I've seen sys.executable before, that's cool

rsheeter: Done

ninja/write_ninja.py
rsheeter: It comes with ninja

src/nanoemoji/nanoemoji.py
```diff
@@ -24,497 +24,135 @@
...
 
-def _write(ufo, ttfont, output_file):
-    logging.info("Writing %s", output_file)
 
-    if os.path.splitext(output_file)[1] == ".ufo":
-        ufo.save(output_file, overwrite=True)
-    else:
-        ttfont.save(output_file)
+def build_dir() -> str:
+    return os.path.abspath(FLAGS.build_dir)
```
anthrotype, line 55: the default build_dir `./build` didn't exist in my repo and it caused a FileNotFoundError when attempting to read from the build.ninja file later.
I think this `build_dir()` function could call `os.makedirs(FLAGS.build_dir, exist_ok=True)` to ensure the requested build directory exists

src/nanoemoji/nanoemoji.py
```diff
@@ -24,497 +24,135 @@
...
-            Component(baseGlyph=base_glyph.name, transformation=Affine2D.identity())
-        )
+    keep_glyph_names = " --"
+    if not FLAGS.keep_glyph_names:
+        keep_glyph_names += "no"
+    keep_glyph_names += "keep_glyph_names"
+    module_rule(
+        "write_font",
+        " --codepoint_csv codepointmap.csv"
+        + " --fea_file features.fea"
```
anthrotype, line 89: if these "codepointmap.csv" and "features.fea" are hard-coded, why do I see them among other flags when I do `nanoemoji --help`?
Are they meant to be overridden by user, if so then you should use the names provided by the caller of nanoemoji CLI.

Also, I would expect that if a codepointmap or features.fea are provided via explicit flags, they are not generated automatically but simply used as they are.

src/nanoemoji/nanoemoji.py
```diff
@@ -24,497 +24,135 @@
...
-        color = next(paint.colors())
-        return palette.index(color)
-
-    elif colr_version == 1:
-        # COLRv1: solid or gradient paint
-        return paint.to_ufo_paint(palette)
 
-    else:
-        raise ValueError(f"Unsupported COLR version: {colr_version}")
+def picosvg(input_svg: str) -> str:
```
anthrotype, line 100: I find the name of this function not very clear, also I fear it may shadow the `picosvg` module name.

src/nanoemoji/nanoemoji.py
```diff
@@ -24,497 +24,135 @@
...
+    nw.build("features.fea", "write_fea", "codepointmap.csv")
+    nw.newline()
 
-    # TODO may wish to nuke 'post' glyph names
 
-    return ufo, ttfont
+def write_font_build(nw: ninja_syntax.Writer, svg_files: Sequence[str]):
+    picosvgs = [picosvg(f) for f in svg_files]
+    nw.build(
+        write_font.output_file(FLAGS.family, FLAGS.output, FLAGS.color_format),
```
anthrotype, line 124: shouldn't the build graph reflect the fact that if ether the codepointmap or the features.fea change, the font should be rebuilt?
this dependency is sort of implicit right now, since the codepoint map and the features derive directly from the filenames of the input svgs. If we allow the user to pass their own codepointmap or features.fea, then write_font should take those as dependencies.

requirements.txt
```diff
@@ -14,6 +14,7 @@ cu2qu==1.6.7              # via ufo2ft
 fonttools[ufo]==4.13.0    # via booleanoperations, cffsubr, compreffor, cu2qu, nanoemoji (setup.py), ufo2ft, ufolib2
 fs==2.4.11                # via fonttools
 lxml==4.5.1               # via nanoemoji (setup.py), picosvg
+ninja==1.10.0.post1
```
anthrotype, line 17: ninja should be added to the setup.py `install_requires` list as well (so that it's pulled automatically when one does `pip install nanoemoji`). The requirements.txt is auto-generated via `pip-compile setup.py`

anthrotype: `pip-compile` script is part of `pip-tools` https://github.com/jazzband/pip-tools

We should add `pip-tools` to the dev-requirements.in file and regenerate `dev-requirements.txt` from it (with `pip-compile dev-requirements.in`)

src/nanoemoji/nanoemoji.py
rsheeter: Fixed

requirements.txt
rsheeter: Fixed.

src/nanoemoji/nanoemoji.py
rsheeter: Updated to picosvg_dest

src/nanoemoji/nanoemoji.py
rsheeter: I like the flags but it's making ninja slightly awkward because it just wants to list them as inputs. I'm going to move write_font to use positional / name pattern based args, remove the flags, and list the codepointmap and feature file as explicit inputs. That should fix the build graph and the messy --help.

rsheeter: Build file looks much cleaner, ty, good feedback.

PR #91, opened 2020-07-29
Image diff support

rsheeter:
Add the ability to render svg, render using updated Skia/FT, and compare the difference for QA purposes.


requirements.txt
```diff
@@ -8,13 +8,16 @@ absl-py==0.9.0            # via nanoemoji (setup.py)
 appdirs==1.4.4            # via fs
 attrs==19.3.0             # via ufolib2
 booleanoperations==0.9.0  # via ufo2ft
+cairosvg==2.4.2
```
anthrotype, line 11: cairosvg and pillow should be added to the setup.py install_requires and then the requirements.txt regenerated via `pip-compile setup.py`, see also https://github.com/googlefonts/nanoemoji/pull/90/commits/64841b49003841eaba1d29de52711441730cfd4c#r462409939

src/nanoemoji/nanoemoji.py
```diff
@@ -94,13 +96,19 @@ def module_rule(mod_name, arg_pattern):
...
+        nw.rule("svg2png", f"cairosvg -H {FLAGS.svg_font_diff_resolution}  -W {FLAGS.svg_font_diff_resolution} -f png -o $out $in")
     nw.newline()
 
 
 def picosvg(input_svg: str) -> str:
     return os.path.join("picosvg", os.path.basename(input_svg))
 
 
+def svg_png(input_svg: str) -> str:
+    return os.path.join("svg_png", os.path.basename(input_svg))
```
anthrotype, line 109: you should change the file extension otherwise you end up saving a .png file (from cairosvg) with an incorrect .svg file extension

requirements.txt
rsheeter: Done and wrote https://googlefonts.github.io/python to try to learn how some of the parts fit together :)

src/nanoemoji/nanoemoji.py
rsheeter: Fixed.

PR #95, opened 2020-07-31
picosvg.svg_meta.ntos no longer rounds; update tests following new picosvg round_floats and absolute coords

anthrotype:
This depends on https://github.com/googlefonts/picosvg/pull/68 which should be merged first

TODO: update requirements to use newly released picosvg after googlefonts/picosvg#68 gets merged


anthrotype:
Note: the changes in the expected test results only change the whitespace formatting of the path.d string, not their content (for readability picosvg adds a space before a command letter).

With latest picosvg all paths are reformatted as we do an explicit SVGShape.round_floats(3) on all the shapes (and not only those that require manipulation by skia or other transformations).

anthrotype:
nanoemoji tests now fail after picosvg is running absolute() on all shapes (it's mostly no-op with the closing "z" that becomes "Z"). Expected filles need to be updated after new picosvg release.

PR #101, opened 2020-08-11
Start to make colr vfs build

rsheeter:
Make nanoemoji do as https://gist.github.com/anthrotype/c11515065a4aa6549f9d5f2dfdcf8f23 does :)


src/nanoemoji/write_variable_font.py
```diff
@@ -0,0 +1,75 @@
...
+
+def only(filter_fn, iterable):
+    it = filter(filter_fn, iterable)
+    result = next(it)
+    assert next(it, None) is None
+    return result
+
+
+def main(argv):
+    ufos = tuple(filter(lambda a: a.endswith(".ufo"), argv[1:]))
```
anthrotype, line 33: i'm seeing a lot of `filter()` lately, is that a new functional programming fad? :) 
doing `tuple(filter(lambda ..., iterable))` is not necessarily faster or more readable than using a list/tuple comprehension or a generator expressions, e.g. `tuple(v for v in iterable if ...)`. It's more a style thing I guess. I don't mind anyway, just want to point this out.

src/nanoemoji/write_variable_font.py
```diff
@@ -0,0 +1,75 @@
...
+
+    axis_names = {a.axisTag: a.name for a in font_config.axes}
+    for master in font_config.masters:
+        ufo = ufoLib2.Font.open(master.output_ufo)
+        ufo.info.styleName = master.style_name
+        location = {axis_names[p.axisTag]: p.position for p in master.position}
+        designspace.addSourceDescriptor(name=master.output_ufo, location=location, font=ufo)
+
+    # build a variable TTFont from the designspace document
+    # TODO what if I wanted a variable CFF?!
```
anthrotype, line 69: if you want a variable CFF2 font you do `ufo2ft.compileVariableCFF2`
https://github.com/googlefonts/ufo2ft/blob/e1c08f6efb07a2fb5202d65a90af94dbd599e7e8/Lib/ufo2ft/__init__.py#L617

src/nanoemoji/write_variable_font.py
rsheeter: Tbh I think whether I write filter vs a comprehension varies based on what I was working on prior to writing it. If I stop and think about it I'd say I prefer the comprehension/generator style for trivial examples like this one.

src/nanoemoji/write_variable_font.py
rsheeter: Is there / could there be a compileVariableCFF?

src/nanoemoji/write_variable_font.py
rsheeter: ...I now want to go change a bunch of recently added filters to use the non-filter style.

tests/minimal_vf/config.toml
```diff
@@ -0,0 +1,20 @@
+output_file = "MinimalVF.ttf"
+color_format = "glyf_colr_1"
+
+[axis.wght]
+name = "Weight"
+default = 300  # apparently ufo2ft fails if default is not at a master. How silly.
```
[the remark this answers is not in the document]
rsheeter: Agreed the silly comment is itself silly, will remove :D 

Thinking "aloud", it would actually be nice to default to letting the compiler pick the default, and if the user chooses to specify then just do it.

For context, I tend to think if you are programmatically compiling a VF you are into the territory of asserting (falsely or not) that you know what you are doing so it would be reasonable for the compiler to either Just Work when you specify an off-master default (maybe you care about non-VF environments and really need that off-master default) or to fail and clearly indicate how to make it work in the error message (specify some flag if you really want this, beware you may inflate the size of your font perhaps?).

src/nanoemoji/write_variable_font.py
rsheeter: [I did]

src/nanoemoji/write_variable_font.py
anthrotype: no, only CFF2 can be variable

src/nanoemoji/nanoemoji.py
```diff
@@ -89,15 +87,40 @@ def _bool_flag(name):
     return flag
 
 
-def write_preamble(nw):
-    def module_rule(mod_name, arg_pattern, rspfile=None, rspfile_content=None):
+def _ufo_rule(master: MasterConfig) -> str:
+    return "write_" + master.style_name.lower() + "_ufo"
```
anthrotype, line 91: could `master.style_name` contain spaces? Should we not `.replace(" ", "")` them? Or will the ninja writer automatically escape them?

src/nanoemoji/nanoemoji.py
```diff
@@ -148,65 +160,64 @@ def module_rule(mod_name, arg_pattern, rspfile=None, rspfile_content=None):
             rspfile="$out.rsp",
             rspfile_content="$in",
         )
+        nw.newline()
+
+    module_rule("write_variable_font", f"--config config.toml $in")
```
anthrotype, line 165: should this rule be written out only if len(masters) > 1 maybe?

src/nanoemoji/write_variable_font.py
```diff
@@ -0,0 +1,78 @@
...
+                if p.axisTag == a.axisTag
+            ),
+        )
+        for a in font_config.axes
+    ]
+    pp.pprint(axis_defs)
+    for axis_def in axis_defs:
+        designspace.addAxisDescriptor(**axis_def)
+
+    axis_names = {a.axisTag: a.name for a in font_config.axes}
```
anthrotype, line 62: in HOI, axisTags may not be unique. Not sure if how future-proof we want to be here.

src/nanoemoji/write_variable_font.py
```diff
@@ -0,0 +1,78 @@
...
+    for master in font_config.masters:
+        ufo = ufoLib2.Font.open(master.output_ufo)
+        ufo.info.styleName = master.style_name
+        location = {axis_names[p.axisTag]: p.position for p in master.position}
```
anthrotype, line 66: in designspaceLib, `axis.name` is guaranteed to be unique unlike axisTag (it's like an internal identifier). Maybe in nanoemoji.config.AxisPosition namedtuple we could use the unique axis "name" instead of axisTag?

src/nanoemoji/write_variable_font.py
anthrotype: I suggest changing the comment here to read "TODO: Use ufo2ft.compileVariableCFF2`

tests/config_test.py
```diff
@@ -0,0 +1,43 @@
...
+    "config_file",
+    [
+        None,  # the default config file
+        "minimal_vf/config.toml",
+        "minimal_static/config.toml",
+    ],
+)
+@pytest.mark.usefixtures("absl_flags")
+def test_read_write_config(config_file):
+    tmp_file = Path(tempfile.mkdtemp())
```
anthrotype, line 32: mkdtemp returns a directory so this variable is more like a `tmp_dir`

tests/minimal_vf/config.toml
```diff
@@ -0,0 +1,20 @@
+output_file = "MinimalVF.ttf"
+
+[axis.wght]
+name = "Weight"
+# ufo2ft fails if default is not at a master.
+default = 300
```
anthrotype, line 6: it's not ufo2ft's fault, it's the way VF works. One of the masters has to be the default. If you don't care which one just use the minimum. Making an arbitrary instance from an initial set of masters and re-inserting this instance as a new master and making it the new default creates a whole new design-space which may not necessarily give identical results as the original set of masters. So the user has to provide the full set of masters, including the default one which is the most important of all.

anthrotype (approved):

src/nanoemoji/config.py
```diff
@@ -0,0 +1,206 @@
...
+from pathlib import Path
+import toml
+from typing import NamedTuple, Tuple, Sequence
+
+
+FLAGS = flags.FLAGS
+
+
+flags.DEFINE_string(
+    "config", str(Path(__file__).parent / "default.toml"), "Config file"
```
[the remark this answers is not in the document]
anthrotype: If you add a data file and want to have it installed alongside the other python modules, you need to pass `include_package_data=True` keyword to setup.py's `setup()`:
https://setuptools.readthedocs.io/en/latest/userguide/datafiles.html

This method works for data files that are tracked by VCS (git) and included in the sdist manifest automatically via setuptools_scm.

src/nanoemoji/config.py
```diff
@@ -0,0 +1,206 @@
...
+)
+
+
+_DEFAULT_FAMILY = "An Emoji Family"
+_DEFAULT_REUSE_TOLERANCE = 0.1
+_DEFAULT_COLOR_FORMAT = "glyf_colr_1"
```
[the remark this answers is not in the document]
rsheeter: These are the defaults for when you have a config file and didn't set a value. Another approach would be to load default then merge the user config on top.

src/nanoemoji/config.py
```diff
@@ -0,0 +1,206 @@
...
+
+    @property
+    def output_format(self):
+        return Path(self.output_file).suffix
+
+
+def _consume(config, key, default=None):
+    if default is None or key in config:
+        return config.pop(key)
+    return default
```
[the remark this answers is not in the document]
rsheeter: I think I was under the impression pop didn't have a default

src/nanoemoji/config.py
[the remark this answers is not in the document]
anthrotype: yeah, the latter sounds clearer to me. We have a default.toml to start with (with all the default values), then a user --config, and finally user CLI options that override both

src/nanoemoji/write_variable_font.py
rsheeter: I fear we'd need much more extensive changes for this tool to support HOI so I'm OK with this not working for now. The sources don't really support it currently.

tests/minimal_vf/config.toml
rsheeter: I don't think the comment implies ufo2ft is doing anything wrong?

anthrotype: It's fine, I misunderstood the "fails" ;)

src/nanoemoji/write_variable_font.py
rsheeter: Done

src/nanoemoji/config.py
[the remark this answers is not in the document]
rsheeter: I'm 99% certain I have other places that use __file__ as well :(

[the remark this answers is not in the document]
anthrotype: nevermind ;)

src/nanoemoji/config.py
```diff
@@ -0,0 +1,206 @@
...
+    axes: Tuple[Axis]
+    masters: Tuple[MasterConfig]
+    source_names: Tuple[str]
+
+    @property
+    def output_format(self):
+        return Path(self.output_file).suffix
+
+
+def _consume(config, key, default=None):
```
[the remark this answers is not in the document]
rsheeter: Now using pop

src/nanoemoji/config.py
```diff
@@ -0,0 +1,206 @@
...
+
+FLAGS = flags.FLAGS
+
+
+flags.DEFINE_string(
+    "config", str(Path(__file__).parent / "default.toml"), "Config file"
+)
+
+
+_DEFAULT_FAMILY = "An Emoji Family"
```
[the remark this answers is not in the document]
rsheeter: Fixed

src/nanoemoji/config.py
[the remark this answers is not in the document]
rsheeter: Still good to fix :)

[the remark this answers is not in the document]
rsheeter: Fixed

src/nanoemoji/write_variable_font.py
rsheeter: IIUC tag is unique until such time as we support HOI which is fundamentally incompatible with our source format currently?

src/nanoemoji/config.py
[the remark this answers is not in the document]
rsheeter: Now we have defaults inlined on the tuple maybe I don't even need default.toml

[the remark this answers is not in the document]
rsheeter: ...actually I think I like the way it's setup now. WDYT?

PR #109, opened 2020-08-19
Use a temporary file to pass content of codepointmap.csv

mavit:
This prevents the error `ninja: fatal: posix_spawn: Argument list too long` from being encountered when a large number of SVGs are to be included.


anthrotype:
thanks, I also noticed this yesterday.
Maybe you could add a comment to explain what an ".rsp" (ninja "response file") is -- something like "works around some OS' max command line length limitations". I confess I had to look it up in the https://ninja-build.org/manual.html

anthrotype:
I'll add the comment, thanks

anthrotype:
it turns out "response files" aren't a ninja-specific thing, more like MSVC feature, since on Windows command line lenght is much shorter than on Unix
https://docs.microsoft.com/en-us/cpp/build/reference/at-specify-a-compiler-response-file?view=vs-2019

mavit:
Indeed.  I found the documentation for this feature to be a bit thin, and I think that's because the target audience is Windows developers who are already familiar with the idiom.

anthrotype:
@mavit I just realised that the idiom `$(<out.rsp)` that you used to read the content of the rsp file into a shell variable doesn't work on my Linux box, whereas it works on the Mac. I think that's a bash-ism, isn't it? Not sure why it isn't working on Linux..

Maybe we should use $(cat out.rsp) for better compat?

Or I'm thinking that write_codepoints.py python module could read from the rsp file that is passed as argument, which would make it work on Windows as well (where cat is not present normally).

We could even adopt the `@out.rsp` syntax that msvc compiler uses in our parsing of command line args..

anthrotype:
another thing I'm wondering is, why would the `posix_spawn: Argument list too long` only occur for the write_codepoints rule, and not also for other rules that take similarly long list of files as positional arguments such as write_font rule?

anthrotype:
Let's continue on https://github.com/googlefonts/nanoemoji/issues/110

mavit:
> another thing I'm wondering is, why would the `posix_spawn: Argument list too long` only occur for the write_codepoints rule, and not also for other rules that take similarly long list of files as positional arguments such as write_font rule?

It's the additional space used by the length of the path multiplied by the number of items.  With 0.3.1:

Fails:
```sh
nanoemoji ~/src/openmoji/color/svg/*.svg
```

Works:
```sh
mkdir svg
cp ~/src/openmoji/color/svg/*.svg svg/
nanoemoji svg/*.svg
```

So, in the future, if enough emojis are added, I think we can expect to need to use a response file for other rules, too.

Issue #113, opened 2020-08-21
Jumbled SVG font


mavit:
I tried to build an SVG font with `nanoemoji --color_format=picosvgz ...`.  When used with Firefox, I found that the images were shuffled into the wrong characters.

```html
<!DOCTYPE html>
<html>
	<body>
		U+1F600 Grinning Face
		
		<p>Default font: 😀</p>
		<p>SVG font: <span style="font-family: An Emoji Family;">😀</span></p>
	</body>
</html>
```

![image](https://user-images.githubusercontent.com/120296/90885166-00c35180-e3a9-11ea-87a9-1bee19dd3fa7.png)

anthrotype:
@mavit Sorry for the radio silence, I'm back on this.
You said "when used in Firefox" -- does it only happen there, or also on other browsers, say, Chrome?

Can you give me a link to the SVG sources and the exact command you run so I can try to reproduce this? Thank you

mavit:
> @mavit Sorry for the radio silence, I'm back on this.
> You said "when used in Firefox" -- does it only happen there, or also on other browsers, say, Chrome?

I tried Firefox because I'm not sure I have anything else that supports SVG fonts.  Fedora's build of Chromium leaves a blank square where the emoji should be.

> Can you give me a link to the SVG sources and the exact command you run so I can try to reproduce this? Thank you

```bash
git clone git://github.com/hfg-gmuend/openmoji.git
nanoemoji --color_format=picosvg ./openmoji/color/svg/*.svg
```

anthrotype:
ah you're right, Chrome doesn't support OT-SVG yet. I think Safari on Mac does, as should do the Adobe design apps.
Anyway, thanks for the reproducer -- I'll look into this soon.

anthrotype:
Ah I see what's going on here. With --color_format=picosvg (or picosvgz) we try to group reused shapes across input SVGs and create a single SVG document in the OT-SVG table so we can reference the shared shapes. When muiltiple glyphs are defined in the same OT-SVG document, these need to be sorted so that members of the same reuse group have contiguous glyph indexes. We do that in `_ensure_groups_grouped_in_glyph_order` function in `nanoemoji.svg` module. 
https://github.com/googlefonts/nanoemoji/blob/f4f36096e9be2b655af530ae5672296a9dc62a8c/src/nanoemoji/svg.py#L152-L165

[...]

PR #125, opened 2020-09-14
[svg] decompile all tables before changing TTFont glyph order

anthrotype:
Fixes https://github.com/googlefonts/nanoemoji/issues/113

When reusing glyph shapes in OT-SVG, we change the order of glyphs so that those belonging to the same group have contiguous glyph indexes.
However, when modifying a TTFont's glyphOrder, we must also make sure that all the tables that reference glyphs by index are also re-compiled with this updated order, otherwise some tables may keep using the old order (e.g. cmap mapping to wrong glyph, or GSUB substituting the wrong glyph etc.).

Once FontTools provides a better API for reordering glyphs we'll make use of that, see https://github.com/fonttools/fonttools/issues/2060

### DIFF
--- a/tests/clocks_rects_picosvg.ttx
+++ b/tests/clocks_rects_picosvg.ttx
@@ -25,15 +25,15 @@
     <cmap_format_4 platformID="0" platEncID="3" language="0">
       <map code="0x20" name=".space"/><!-- SPACE -->
       <map code="0xe000" name="e000"/><!-- ???? -->
-      <map code="0xe001" name="e002"/><!-- ???? -->
-      <map code="0xe002" name="e001"/><!-- ???? -->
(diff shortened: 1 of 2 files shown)


tests/clocks_rects_picosvg.ttx
```diff
@@ -25,15 +25,15 @@
     <cmap_format_4 platformID="0" platEncID="3" language="0">
       <map code="0x20" name=".space"/><!-- SPACE -->
       <map code="0xe000" name="e000"/><!-- ???? -->
-      <map code="0xe001" name="e002"/><!-- ???? -->
```
anthrotype, line 28: funny how this bug was literally under our 👃

PR #164, opened 2020-10-24
Update to latest COLR spec

rsheeter:
Creating a PR because I like being able to refer to it

Requires picosvg and FontTools updates that are on branches as of 10/24/2020 EDIT: now released


src/nanoemoji/colors.py
```diff
@@ -229,3 +229,16 @@ def to_ufo_color(self):
 
     def opaque(self):
         return self._replace(alpha=1.0)
+
+    def to_string(self):
```
anthrotype, line 233: this could even be the magic `__str__` method if you like, so str(color) or print(color) just work

src/nanoemoji/paint.py
```diff
@@ -111,21 +143,96 @@ class PaintRadialGradient(Paint):
...
+            "format": self.format,
+            "mode": self.mode.name.lower(),
+            "source": self.source.to_ufo_paint(colors),
+            "backdrop": self.backdrop.to_ufo_paint(colors),
+        }
+        return paint
+
+@dataclasses.dataclass(frozen=True)
+class PaintColrLayers(Paint):
+    format: ClassVar[int] = 8
```
anthrotype, line 238: no newline at EOF, my vim always forces me to add one

src/nanoemoji/write_font.py
```diff
@@ -310,23 +331,95 @@ def _glyf_ufo(ufo, color_glyphs):
...
+        )
+
+        i = MAX_LAYER_V1_COUNT - 1
+        colr_layers[i:] = [_build_composite_layer(colr_layers[i:])]
+
+    return colr_layers
+
+
+def _build_composite_layer(layers: Sequence[Mapping[str, Any]]) -> Mapping[str, Any]:
+    """Construct PaintComposite binary tree from list of UFO color layers.
```
anthrotype, line 400: I guess now we could use a flatter tree of PaintColrLayers. Maybe in a separate PR if you prefer.

src/nanoemoji/write_glyphgraph.py
```diff
@@ -0,0 +1,189 @@
...
+        return "_".join(id_parts)
+    if paint.Format == 7:
+        id_parts = (
+            "Composite",
+            _paint_node_id(palette, paint.SourcePaint),
+            paint.CompositeMode.name,
+            _paint_node_id(palette, paint.BackdropPaint),
+            )
+        return "_".join(id_parts)
+    raise NotImplementedError(f"id for format {paint.Format} ({dir(paint)})")
```
anthrotype, line 136: we need an elif paint.Format == 8 for making a digraph with PaintColrLayers

rsheeter: Oh I have that; must just not have uploaded it

src/nanoemoji/colors.py
rsheeter: I waffled ... I kind of like the tuple print

src/nanoemoji/paint.py
rsheeter: black is removing it when I add one

src/nanoemoji/write_font.py
rsheeter: +1 to flatter and separate PR

src/nanoemoji/color_glyph.py
```diff
@@ -134,74 +122,89 @@ def _parse_linear_gradient(grad_el, shape_bbox, view_box, upem):
...
-
-    gradient = {
-        "c0": c0,
-        "c1": c1,
-        "r0": r0,
-        "r1": r1,
-        "affine2x2": (affine2x2[:4] if affine2x2 != Affine2D.identity() else None),
-    }
+    # can't have skew/rotation here: upem, view_box, shape_bbox are all rectangular
+    assert transform[1:3] == (0, 0)
```
rsheeter, line 148: Suggest moving comment to message so if it fails we get a clear complaint

src/nanoemoji/svg.py
```diff
@@ -67,65 +95,224 @@ def _glyph_groups(color_glyphs: Sequence[ColorGlyph]) -> Tuple[Tuple[str, ...]]:
...
+    svg_defs: etree.Element,
+    paint: PaintRadialGradient,
+    transform: Affine2D = Affine2D.identity(),
+) -> str:
+    gradient = etree.SubElement(svg_defs, "radialGradient")
+    gradient_id = gradient.attrib["id"] = f"g{len(svg_defs)}"
+
+    fx, fy = paint.c0
+    cx, cy = paint.c1
+    if fx != cx or fy != cy:
```
rsheeter, line 223: Optional, I wonder if paint.c0 != paint.c1 might be marginally easier to read?

rsheeter:
LGTM

anthrotype (approved):

PR #184, opened 2020-11-14
Increase reuse tolerance default to 0.2

rsheeter:
Helps with https://github.com/googlefonts/picosvg/issues/138

### DIFF
--- a/src/nanoemoji/color_glyph.py
+++ b/src/nanoemoji/color_glyph.py
@@ -40,9 +40,10 @@ import ufoLib2
 FLAGS = flags.FLAGS
 
 
+# Bump to 0.2 per https://github.com/googlefonts/picosvg/issues/138
 flags.DEFINE_float(
-    "reuse_tolerence",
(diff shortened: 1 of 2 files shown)


anthrotype:
I don't know if it's a good idea to increase the default reuse tolerance to 0.2..
For a 128-unit wide SVG viewBox (such as noto-emoji's), when scaled to a font UPEM of 1024, the tolerance of 0.2 becomes 0.2 * 8 = 1.6 font units, which could become visible.

I tried to visualize the test case in picosvg where you increased the tolerance to 0.2 (https://github.com/googlefonts/picosvg/pull/139) and that's how it looks like:

image

(source: https://codepen.io/anthrotype/pen/WNxWoeO)

rsheeter:
> I think it looks too different to be acceptable

I agree for text but I wonder if it's ok for emoji. That said, it would clearly be better if we just computed a transform that didn't have that degree of error.

PR #186, opened 2020-11-19
add flag to --ignore_reuse_errors

anthrotype:
This allows to continue compilation with a warning when picosvg fails to compute `affine_between` two paths that have the same normalized path.

see https://github.com/googlefonts/picosvg/issues/138

### DIFF
--- a/src/nanoemoji/color_glyph.py
+++ b/src/nanoemoji/color_glyph.py
@@ -50,7 +50,7 @@ flags.DEFINE_float(
 # https://github.com/googlefonts/picosvg/issues/138
 flags.DEFINE_bool(
     "ignore_reuse_error",
-    False,
+    True,
     "Whether to fail or continue with a warning when picosvg cannot compute "
     "affine between paths that normalize the same."
 )
--- a/src/nanoemoji/color_glyph.py
+++ b/src/nanoemoji/color_glyph.py
@@ -47,6 +47,13 @@ flags.DEFINE_float(
     " Normalized shapes snap to whole multiples of tolerance;"
     " choice of a value where 1/tolerance is an int recommended",
 )
+# https://github.com/googlefonts/picosvg/issues/138
+flags.DEFINE_bool(
+    "ignore_reuse_error",
+    False,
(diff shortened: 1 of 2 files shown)


src/nanoemoji/color_glyph.py
```diff
@@ -278,12 +285,25 @@ def _painted_layers(
...
+                if FLAGS.ignore_reuse_error:
+                    logging.warning(error_msg)
+                else:
+                    raise ValueError(error_msg)
+
+        if success:
+            yield PaintedLayer(paint, paths[0].d, transforms)
+        else:
+            for path in paths:
+                yield PaintedLayer(paint, path.d)
```
rsheeter, line 306: I don't think you want to drop all reuse when a subset fail to compute a transform. Maybe tweak to something roughly like:

```python
# looping twice should be fine, small #s
no_transform = [p for p,t in zip(...) if t is None]
# log or fail as per --ignore_reuse_error; can now report multiple failures in ValueError which is nice
yield PaintedLayer(paint, paths[0].d, [t for t in ... if t is not None])
```

src/nanoemoji/color_glyph.py
```diff
@@ -47,6 +47,13 @@
     " Normalized shapes snap to whole multiples of tolerance;"
     " choice of a value where 1/tolerance is an int recommended",
 )
+# https://github.com/googlefonts/picosvg/issues/138
+flags.DEFINE_bool(
+    "ignore_reuse_error",
+    False,
```
rsheeter, line 53: I actually think you want True; it's nice when things work

rsheeter (approved):

src/nanoemoji/color_glyph.py
rsheeter: Updated

src/nanoemoji/color_glyph.py
rsheeter: In retrospect that changes the layer order, so you actually need to find runs where it worked

rsheeter: Leaving that for a followon pr because I want to test to see how widespread the reuse dropping is

PR #187, opened 2020-11-20
fix reused component transform matrix for font space

anthrotype:
the base glyph is drawn in font space, whereas the transforms in `PaintedLayer.reuses` are in SVG space.
We must adapt the transform such that it first undoes the svg-to-upem transform, applies itself and finally restores svg-to-upem.

Before:

image

After:

image


src/nanoemoji/write_font.py
```diff
@@ -259,11 +259,10 @@ def _create_glyph(color_glyph: ColorGlyph, painted_layer: PaintedLayer) -> Glyph
...
         for transform in painted_layer.reuses:
-            # We already redrew the component into font space, don't redo it
-            # scale x/y translation and flip y movement to match font space
-            transform = transform._replace(
-                e=transform.e * svg_units_to_font_units.a,
-                f=transform.f * svg_units_to_font_units.d,
+            # We already drew the component into font space; transform is in SVG space
+            transform = Affine2D.product(
+                svg_units_to_font_units.inverse(),
+                Affine2D.product(transform, svg_units_to_font_units),
```
rsheeter, line 265: this seems like a place we should use compose_ltr rather than nested products

tests/write_font_test.py
```diff
@@ -104,6 +104,7 @@ def test_keep_glyph_names(svgs, color_format, keep_glyph_names):
         ),
         # keep single-component composites if component reused by more than one glyph
         (("one_rect.svg", "one_rect.svg"), "reused_rect_glyf.ttx", "glyf", ".ttf"),
+        (("reused_shape.svg",), "reused_shape_glyf.ttx", "glyf", ".ttf"),
```
rsheeter, line 107: suggest adding a comment to explain what is being tested for

rsheeter (approved):

tests/write_font_test.py
rsheeter: Added

src/nanoemoji/write_font.py
rsheeter: Done

Issue #200, opened 2020-11-26
Allow discarding content outside viewbox


rsheeter:
Noto Emoji and possibly others have content that is entirely outside the viewbox that we should just discard (provided there is no id so it can't be referenced in). When rendering the svg to a bitmap for CBDT this content was ignored.

May want this to be a config option.

A quick sanity check of Noto (run on picosvgs) found quite a few:

```shell
emoji_u1f306.svg (-140.0, 4.0, 120.0, 120.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-136.0, 32.0, 28.0, 76.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-100.0, 12.0, 76.0, 104.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-140.0, 56.0, 120.0, 60.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-140.0, 90.0, 120.0, 34.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-140.0, 4.0, 120.0, 120.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-80.0, 82.0, 6.0, 6.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-64.0, 78.0, 6.0, 6.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-38.0, 66.0, 6.0, 6.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-32.0, 78.0, 6.0, 6.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-108.0, 64.0, 6.0, 6.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-122.0, 72.0, 6.0, 6.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-110.0, 82.0, 6.0, 6.0) outside of Rect(x=0, y=0, w=128, h=128)
emoji_u1f306.svg (-130.0, 48.0, 6.0, 6.0) outside of Rect(x=0, y=0, w=128, h=128)
[...]

PR #201, opened 2020-11-26
Rough in wiring to optionally destroy content outside the viewbox

rsheeter:
First pass at a fix for #200. Requires https://github.com/googlefonts/picosvg/pull/146.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ fs==2.4.11                # via fonttools
 importlib-resources==3.3.0  # via cffsubr
 lxml==4.6.1               # via nanoemoji (setup.py), picosvg
 ninja==1.10.0.post2       # via nanoemoji (setup.py)
-picosvg==0.7.2            # via nanoemoji (setup.py)
+picosvg==0.7.3            # via nanoemoji (setup.py)
 pillow==8.0.1             # via nanoemoji (setup.py)
 pyclipper==1.2.0          # via booleanoperations
 pytz==2020.4              # via fs
--- a/setup.py
+++ b/setup.py
@@ -29,7 +29,7 @@ setup(
         "importlib_resources>=3.3.0",
         "lxml>=4.0",
         "ninja>=1.10.0.post1",
-        "picosvg>=0.7.1",
+        "picosvg>=0.7.3",
         "pillow>=7.2.0",
         "regex>=2020.4.4",
         "toml>=0.10.1",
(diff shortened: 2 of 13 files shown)


rsheeter:
I need the fast CI here too lol

anthrotype:
> I need the fast CI here too lol

I know.. Today I finished touching up the picosvg CI setup. I think I like it now. Initially I had two separate workflows (distinct yaml configs for the tests and the deployment), but now I merged them in to a single worfklow. This way when we `git push --follow-tags` to make a release, it first runs all the tests (the same as with regular pushes) and only if all tests pass, it creates the Github Release and uploads the distribution packages to PyPI. Before the two test and deploy worfklows were independent from one another, and it might happen that the wheel builds but the tests fail and you end up publishing a broken package.

I will update the googlefonts/python docs tomorrow so you can try doing the same for nanoemoji.

anthrotype (approved):
LGTM

rsheeter:
Seems to help rendering considerably to not have bizarre nonsense outside the rendering area. Curiously it seems like the emoji I was testing didn't have crazy stuff but others in the font did and that was enough to make things go awry.

*Before, canvas 3705.4 x 1080.8*

```xml
  <head>
    <xMin value="-9824"/>
    <yMin value="-2408"/>
    <xMax value="16747"/>
    <yMax value="6238"/>

  <hhea>
    <minLeftSideBearing value="-9824"/>
    <minRightSideBearing value="-15723"/>
    <xMaxExtent value="16747"/>
```

image

*After, canvas 512.6 x 128.0*

```xml
  <head>
    <xMin value="-2"/>
    <yMin value="0"/>
    <xMax value="1027"/>
    <yMax value="1024"/>

  <hhea>
    <minLeftSideBearing value="-2"/>
    <minRightSideBearing value="-998"/>
    <xMaxExtent value="1027"/>
```

image

rsheeter:
Made some of our files smaller too which is nice :)

```
 fonts/noto-cff2_colr_1.otf               | Bin 3002612 -> 2976012 bytes
 fonts/noto-cff_colr_1.otf                | Bin 3449160 -> 3416684 bytes
 fonts/noto-glyf_colr_1.ttf               | Bin 3826288 -> 3778192 bytes
 fonts/noto-picosvg.ttf                   | Bin 18771972 -> 18561436 bytes
 fonts/noto-picosvgz.ttf                  | Bin 4557516 -> 4498080 bytes
```

anthrotype:
Wow `3826288 - 3778192 = 48,096`.. is that 48 Kb of junk?

PR #209, opened 2020-12-01
Diff touchups

rsheeter:
Playing with tidying svgs before resvg

### DIFF
--- a/src/nanoemoji/write_tidysvg.py
+++ b/src/nanoemoji/write_tidysvg.py
@@ -1,0 +1,32 @@
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


anthrotype:
I don't think we should worry too much about resvg warning about those invisible paths with only a moveto in the original SVGs. They just won't be drawn, that's fine. But we do need to ensure that we are not encoding them in the generated color fonts. Does picosvg already get rid of them? If so then we're good.

+1 on also showing the diff pixels

src/nanoemoji/write_tidysvg.py
```diff
@@ -0,0 +1,32 @@
...
+https://github.com/googlefonts/nanoemoji/issues/208
+"""
+from absl import app
+from picosvg.svg import SVG
+
+
+def main(argv):
+    src_file, dest_file = argv[1:]
+    svg = SVG.parse(src_file)
+    svg.absolute(inplace=True).remove_unpainted_shapes(inplace=True)
```
anthrotype, line 26: i think it would be better to have resvg render the original SVGs are they are, without introducing any changes. If we want we can fix up the sources as needed. The purpose of the comparison is check that the output of our build pipeline renders the same as the inputs would render, if we are also modifying the input (even with a minor relatively change like this) then we are no longer comparing original vs generated.

PR #222, opened 2020-12-08
Allow CLI flags to override (default) config

### LINKED ISSUE #220: passing CLI flags no longer works after --config introduced
`nanoemoji.write_font` used to have some flags which have been moved to config file following #101, notably:

https://github.com/googlefonts/nanoemoji/pull/101/files#diff-f00e3bf2ff02359126248ec7ef513290cd79523eee44e4e23f9ae3a4a10df769L146-L164

I would be nice to still be able to pass them on the nanoemoji command-line. Sometimes one has a bunch of SVGs and wants to quickly try make a font. Right now to one has to write a .toml config file and pass it as --config option.

Ideally the CLI options would override any --config that was passed.

anthrotype:
Fixes #220 

Also fixes use of misspelled `--reuse_tolerence` (while the config's `reuse_tolerance` was being ignored).

A negative value to `--reuse_tolerance` option means completely disabling in-glyph shape reuse.

### DIFF
--- a/src/nanoemoji/config.py
+++ b/src/nanoemoji/config.py
@@ -23,14 +23,47 @@ from pathlib import Path
 import toml
 from typing import Any, Iterable, MutableMapping, NamedTuple, Optional, Tuple, Sequence
 
+from nanoemoji.write_font import _COLOR_FORMAT_GENERATORS
+
 
 FLAGS = flags.FLAGS
 
 
-_DEFAULT_CONFIG = "_default.toml"
+_DEFAULT_CONFIG_FILE = "_default.toml"
 
-
-flags.DEFINE_string("config", _DEFAULT_CONFIG, "Config file")
+# we use None as a sentinel for flag not set; FontConfig class has the actual defaults.
+# CLI flags override config file (which overrides default FontConfig).
+flags.DEFINE_string("config", None, "Config file")
+flags.DEFINE_integer("upem", None, "Units per em.")
+flags.DEFINE_string("family", None, "Family name.")
+flags.DEFINE_string("output_file", None, "Output filename.")
+flags.DEFINE_enum(
+    "color_format",
+    None,
+    sorted(_COLOR_FORMAT_GENERATORS.keys()),
+    "Type of font to generate.",
+)
+flags.DEFINE_enum(
+    "output", None, ["ufo", "font"], "Whether to output a font binary or a UFO"
+)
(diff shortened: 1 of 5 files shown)


src/nanoemoji/config.py
```diff
@@ -23,14 +23,47 @@
 import toml
 from typing import Any, Iterable, MutableMapping, NamedTuple, Optional, Tuple, Sequence
 
-
-FLAGS = flags.FLAGS
+from nanoemoji import write_font
```
rsheeter, line 26: This seems like an inappropriate import to me, I think we should avoid it and that will conveniently avoid the problem with cycles

src/nanoemoji/config.py
```diff
@@ -23,14 +23,47 @@
...
+# we use None as a sentinel for flag not set; FontConfig class has the actual defaults.
+# CLI flags override config file (which overrides default FontConfig).
+flags.DEFINE_string("config", None, "Config file")
+flags.DEFINE_integer("upem", None, "Units per em.")
+flags.DEFINE_string("family", None, "Family name.")
+flags.DEFINE_string("output_file", None, "Output filename.")
+flags.DEFINE_enum(
+    "color_format",
+    None,
+    sorted(write_font._COLOR_FORMAT_GENERATORS.keys()),
```
rsheeter, line 43: I suggest you enumerate color formats in config.py, import that into write_font, and use it as the key to _COLOR_FORMAT_GENERATORS

tests/write_font_test.py
```diff
@@ -93,6 +93,12 @@ def test_keep_glyph_names(svgs, color_format, keep_glyph_names):
             "clocks_picosvg.ttx",
             {"color_format": "picosvg"},
         ),
+        # passing a tiny --reuse_tolerance effectively disables shape reuse
```
rsheeter, line 96: Maybe we should have a value that actually disables it, such as -1

rsheeter (approved):
LG except I think we should avoid the reference cycle

tests/write_font_test.py
anthrotype: Done

src/nanoemoji/config.py
anthrotype: Done

src/nanoemoji/config.py
anthrotype: removed the cycle

PR #225, opened 2021-01-03
build tree of PaintColrLayers when len(paints) > 255

anthrotype:
The BASKET emoji U+1F9FA (which contains 365 layers) ends up 424 bytes more compact than when we use a tree of SRC_OVER PaintComposites.


anthrotype:
the tests fail for unrelated reasons (in picosvg we changed the starting point of the two arcs that approximate ellipses to match the SVG spec). If we had pinned the test requirements this would not have happened -- but I guess this is a "feature", not a bug :)

anthrotype:
It just occurred to me that we may want instead to move this code down to fontTools.colorLib where we buildPaintColrLayers, so that it's done automatically any time a list of Paints exceeds 255.

src/nanoemoji/util.py
```diff
@@ -56,3 +56,34 @@ def fs_root() -> Path:
 def rel(from_path: Path, to_path: Path) -> Path:
     # relative_to(A,B) doesn't like it if B doesn't start with A
     return Path(os.path.relpath(str(to_path.resolve()), str(from_path.resolve())))
+
+
+def build_n_ary_tree(leaves, n):
+    """Build N-ary tree from sequence of leaf nodes.
+
+    Return a list of lists where each non-leaf node is a list containing
+    max n nodes.
```
rsheeter, line 65: your asserts preclude a tree of no nodes being an empty list ... why would that be wrong?

rsheeter:
> we may want instead to move this code down to fontTools.colorLib where we buildPaintColrLayers

+1

anthrotype:
I moved the code to https://github.com/fonttools/fonttools/pull/2153

PR #226, opened 2021-01-11
Update to FontTools 4.18.2 to get latest Paint formats

rsheeter:


rsheeter:
@anthrotype I'm failing to spot why we have different sets of <pt> on https://github.com/fonttools/fonttools/releases/tag/4.18.2?

anthrotype:
The cause of the ttx diffs is not fonttools but picosvg, where we changed the starting position of the two arcs approximating ellipse and circle shapes

anthrotype:
https://github.com/googlefonts/nanoemoji/pull/225#issuecomment-753658950

anthrotype:
When multiple dependencies get updated at the same time and you don't control the timing of the update because they just happen upstream and we let pip install the latest, then it becomes a little bit harder to track why unit tests tart

rsheeter:
Oh that makes sense; I think I should update picosvg min version then. It's a bit silly to have tests that fail or succeed depending on what version of picosvg you have when setup claims both valid.

anthrotype:
Yes we should update minimum picosvg as well

anthrotype:
Sorry I inadvertently closed the PR. I should probably go back to sleep 😴

rsheeter:
lol does it look ok to merge? :)

anthrotype (approved):

PR #227, opened 2021-01-12
Fix --ignore_reuse_error

rsheeter:
Make it consistent with other config flags; seems to be causing my color-fonts updates to fail because nanoemoji helpfully passes --noignore_reuse_errors along when the flag isn't specified.


rsheeter:
@anthrotype could I impose on you for a post-merge review? - I see I got a little ahead of myself

anthrotype:
LGTM

PR #233, opened 2021-02-02
Utility to count overhead due to variation indices

rsheeter:
We've talked from time to time about overhead due to making *everything* variable. Raph and Just also raised this. First draft of a utility to count. 

Sample result:

```shell
$ python -m nanoemoji.vf_overhead ../color-fonts/fonts/noto-glyf_colr_1.ttf
633476 / 1340960 bytes (47.2%) of the COLR table are variation indices
633476 / 3784340 bytes (16.7%) of the font file are variation indices

$ python -m nanoemoji.vf_overhead ../color-fonts/fonts/noto-cff_colr_1.otf 
631884 / 1340254 bytes (47.1%) of the COLR table are variation indices
631884 / 3416684 bytes (18.5%) of the font file are variation indices
```

Should check impact on woff2 as well (ty @drott for the reminder). Edit: did check, see https://github.com/googlefonts/colr-gradients-spec/issues/233 for better numbers (actually built font with/without varidx).

### DIFF
--- a/src/nanoemoji/vf_overhead.py
+++ b/src/nanoemoji/vf_overhead.py
@@ -1,0 +1,91 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helps estimate overhead due to VF offsets."""
+
+from absl import app
+from fontTools.colorLib.builder import LayerV1ListBuilder
+from fontTools import ttLib
+from fontTools.ttLib.tables import otTables as ot
+from nanoemoji import colr_traverse
+from pathlib import Path
+
+# VarIdx costs 4 bytes in every Var<something>
+# ColorIndex costs 4: alpha
+# ColorStop costs 8: 4 (stopOffset) + 4 (ColorIndex)
+# ColorLine costs #stops * 8 : ColorStop
+# Affine2x3 costs 24: 4 per field
+def _overhead_color_line(color_line):
+    return 8 * len(color_line.ColorStop)
+
+def _overhead_gradient(color_line, num_fields):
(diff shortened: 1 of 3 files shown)


src/nanoemoji/vf_overhead.py
```diff
@@ -0,0 +1,91 @@
...
+
+    def _callback(paint):
+        overhead = 0
+        paint_tuple = list_builder._paint_tuple(paint)
+        seen = paint_tuple in visited
+        if not seen:
+            visited.add(paint_tuple)
+            overhead = _OVERHEAD[paint.Format](paint)
+        #print("Visit", paint.Format, "Seen?", seen, "VarOverhead", overhead)
+        total_overhead[0] += overhead
```
anthrotype, line 66: are you using a list just to be able to assign to a variable in the outer scope?
you can annotate it with `nonlocal` to achive that, see https://docs.python.org/3/reference/simple_stmts.html#the-nonlocal-statement

rsheeter: > are you using a list just to be able to assign to a variable in the outer scope?

 Yes :) Good tip, ty

src/nanoemoji/vf_overhead.py
```diff
@@ -0,0 +1,91 @@
...
+    total_overhead = 0
+    visited = set()
+    list_builder = LayerV1ListBuilder()
+
+    def _callback(paint):
+        overhead = 0
+        paint_tuple = list_builder._paint_tuple(paint)
+        seen = paint_tuple in visited
+        if not seen:
+            visited.add(paint_tuple)
```
anthrotype, line 63: I'm not sure if it's correct to only visit unique paints hashed by value... The `_paint_tuple` method returns a hashable tuple of the data inside a Paint object (and its children). But in the same COLR table there could be multiple instances of a given subgraph identified by the same paint tuple. IIUC, you're only counting them the first time each unique paint appear and skipping the duplicates. The arguable assumption being that there are no actua duplicates (i.e. distinct paints or clusters with same content), but only duplicate traversals of the same node pointed to by multiple edges.. Does that make sense?

rsheeter: My underlying assumption was ensuring that duplicates don't occur was something the FontTools writer does when emitting content. Is that not true or only partially true?

Hm. Probably I should just close this PR; the test of actually compiling a font with fonttools after removing all the VarTs is the authoritative answer.

anthrotype: > something the FontTools writer does when emitting content

that's true.

> compiling a font with fonttools after removing all the VarTs is the authoritative answer

sgtm

PR #234, opened 2021-02-05
renumber Paint format numbers in tests, rename Paint.Format=>PaintFormat

anthrotype:
format numbers changed after the introduction of PaintSweepGradient, and the non-variable Paint formats.
Tests need updating.

This PR depends on fonttools >= 4.20.0

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,7 @@ setup(
     install_requires=[
         "absl-py>=0.9.0",
         "dataclasses>=0.8; python_version < '3.7'",
-        "fonttools[ufo]>=4.19.0",
+        "fonttools[ufo]>=4.20.0",
         "importlib_resources>=3.3.0; python_version < '3.9'",
         "lxml>=4.0",
         "ninja>=1.10.0.post1",
--- a/src/nanoemoji/paint.py
+++ b/src/nanoemoji/paint.py
@@ -92,23 +92,27 @@ class PaintSolid(Paint):
 
     def to_ufo_paint(self, colors):
         return {
-            "format": self.format,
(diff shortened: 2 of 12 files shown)


rsheeter (approved):

rsheeter (approved):

rsheeter:
> This PR depends on fonttools/fonttools#2175 being merged first (until then tests will fail)

bump the min FontTools required in this PR perhaps? (once FT releases)

anthrotype:
yeah, that's why it's marked draft, because we first need to release new FT

anthrotype:
I pushed a couple of commits to make nanoemoji work with the colrv1_static branch of https://github.com/fonttools/fonttools/pull/2182

There are still 4 tests that are failing even after that. 2 relate to the fact that fonttools is no longer unpacking single-item PaintColrLayers as it was doing before (I commented on that in https://github.com/fonttools/fonttools/pull/2181#discussion_r575359256).

The other 2 failing tests are suspicious and need to be looked at, preferably before merging https://github.com/fonttools/fonttools/pull/2182. They show off-by-one difference in some PaintLinearGradient point coordinates which are unexpected and must be investigate to avoid regressions.

rsheeter:
> They show off-by-one difference in some PaintLinearGradient point coordinates which are unexpected and must be investigate to avoid regressions.

Good catch, located a missing otRound. https://github.com/fonttools/fonttools/pull/2181/commits/5316ae4b8cef78dc8af3479ee6fa87bb95cc4a6c.

All nanoemoji tests now pass for me when using the latest FontTools from https://github.com/fonttools/fonttools/pull/2182 [which sits on top of https://github.com/fonttools/fonttools/pull/2181].

PR #240, opened 2021-02-22
also run tests on windows-latest python3.9

anthrotype:


anthrotype:
I finally managed to reproduce the issue reported by @paullinnerud, see:

https://github.com/googlefonts/nanoemoji/pull/240/checks?check_run_id=1952286290#step:5:160

> ValueError: Bad filename >; unable to extract codepoints

The problem is that nanoemoji-generated ninja rules are making extensive use of the special `>` character for dumping stdout to a file, but that doesn't seem to work on Windows. I think ninja uses CreateProcess API to invoke commands on Windows, so we can't use bash specific notation, unless we wrap in a `bash -c "..."` but then we make nanoemoji dependant on bash.exe.

Need to think how to best resolve this. In the meantime, I recommend that if you wish to use nanoemoji on Windows, you try running it from inside the [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/about).

anthrotype:
i guess the only solution is to avoid using `>` altogether and instead change picosvg and the other nanomoji sub-commands to have an `-o` command line option to specify the output file name, instead of relying on redirecting standard output to a file.

anthrotype:
All tests finally pass on Windows too! Huzzah! 🕺

.github/workflows/ci.yml
```diff
@@ -22,17 +22,21 @@ jobs:
...
     strategy:
       matrix:
-        python-version: [3.6, 3.7, 3.8, 3.9]
-        platform: [ubuntu-latest]
+        config:
+          - {platform: ubuntu-latest, python-version: 3.6}
+          - {platform: ubuntu-latest, python-version: 3.7}
+          - {platform: ubuntu-latest, python-version: 3.8}
+          - {platform: ubuntu-latest, python-version: 3.9}
+          - {platform: windows-latest, python-version: 3.9}
```
anthrotype, line 33: I think it's fine to only test the latest stabe python on Windows, because win users have to download/install it anyway so they end up using whatever is the latest. Also I wouldn't want to slow down the CI too much.

rsheeter (approved):
LG, ty

paullinnerud:
Thanks, I verified it works on my Windows system.

PR #248, opened 2021-03-03
[WIP]fix: avoid scaling the viewbox width to cause glyph distortion

yisibl:
In this PR, the glyph width is allowed to exceed upem to avoid distortion.

- [ ] Need to add a flag to turn on this feature (monospaced or not)

(The blurring in after should be a Chrome bug.)

## Test svg
```svg
<svg width="400" height="100" viewBox="0 0 400 100" xmlns="http://www.w3.org/2000/svg">
    <path fill="#E45019" fill-rule="nonzero" d="M299.097.91l101.39 50.14-101.531 49.86.056-40.385-288.77-.394-9.756-.013.027-19.23 9.756.012 288.77.395z" />
</svg>
```

### DIFF
--- a/src/nanoemoji/color_glyph.py
+++ b/src/nanoemoji/color_glyph.py
@@ -36,11 +36,17 @@ from picosvg.svg_types import SVGPath, SVGLinearGradient, SVGRadialGradient
 from typing import Generator, NamedTuple, Optional, Sequence, Tuple
 import ufoLib2
 
-
 def _scale_viewbox_to_emsquare(view_box: Rect, upem: int) -> Tuple[float, float]:
-    # scale to font upem
-    return (upem / view_box.w, upem / view_box.h)
+    x_scale = upem / view_box.w
+    y_scale = upem / view_box.h
+    aspect_ratio = view_box.w / view_box.h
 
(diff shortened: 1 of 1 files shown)


src/nanoemoji/color_glyph.py
```diff
@@ -36,11 +36,16 @@
...
     # scale to font upem
-    return (upem / view_box.w, upem / view_box.h)
-
+    view_box_w = upem / view_box.w
+    view_box_h = upem / view_box.h
+
+    # upem = 1024
+    # Don't zoom when the viewbox width is greater than upem.
+    if view_box.w > upem:
+        view_box_w = 1.0
```
anthrotype, line 47: but here you are not maintaining the aspect ratio. You're only handling the width independently of the height, and only when viewbox.w exceeds the upem (which is not always 1024 btw).
If you scale the width and height differently you won't preserve the aspect ratio.

yisibl: Yes, I haven't figured out how to keep the aspect ratio while making it overflow. How do we scale the graphics while maintaining the aspect ratio? Do you have any advice?

anthrotype:
It should be fixed with #255, thank you.

Issue #238, opened 2021-02-16
Compute bbox correctly for static


rsheeter:
Whatever answer https://github.com/googlefonts/colr-gradients-spec/issues/251 lands on we need to compute the bbox correctly.

This may be awkward under variation, esp with `PaintRotate`.

See https://github.com/googlefonts/nanoemoji/blob/master/src/nanoemoji/write_font.py#L260.

drott:
CC @bungeman

rsheeter:
Context for future self: we need to update to account for the DAG; after #252 it should be correct for what nanoemoji actually generates but not for arbitrary graphs.

rsheeter:
I believe the new ClipBox mechanism does this correctly.

PR #250, opened 2021-03-04
Add user configured ascent, descent, transform

### LINKED ISSUE #2: feature request: option to shift vertically the "baseline" of SVGs
Currently we map the svg viewbox (x, y) point to (0, UPEM) in font space.
We scale the viewbox width/height so they fit the units-per-em, and flip the Y-axis (svg goes downwards, font go upwards) and translate the origin y += UPEM.

E.g. if viewbox is (0, 0, 10, 10) and font UPEM is 1000 we end up with a transform from svg to font space like this: (100, 0, 0, -100, 0, 1000). So, point (0, 0) in svg space maps to (0, 1000) in font space, point (0, 10) in svg maps to (0, 0) in font, point (10, 10) in svg maps to (1000, 0), etc.

This may be good for now, however positioning the origin so high up by an amount equal to the UPEM may not work for all designs.
Even for noto or twemoji I think we end up with emojis that look don't look centered next to other text.

Maybe nanoemoji could have a commandline option to customize the vertical placement of the SVG in the font space. Or some option that allows to set the imaginary "baseline" of the emojis (i.e. what would become after the transformation the (0,0) in font units).

I haven't fully thought of a UX for this but I think it could be useful to have.

rsheeter:
WIP. Meant to fix #2.

Built on https://github.com/googlefonts/nanoemoji/pull/249.

As we parse svg style transform strings things like `nanoemoji emoji_u2702.svg --transform "translate(-512,-512) scale(1.17) translate(512, 512)"` become possible on CLI or in config file.

For reference, values from a test run of the Noto emoji_builder:

```
Loaded font 'NotoColorEmoji.tmpl.ttf'.
Font metrics: upem=2048 ascent=1900 descent=500.

Looking for images matching 'build/compressed_pngs/emoji_u*.png'.
Found images for 1 characters in 'build/compressed_pngs/emoji_u*.png'.
Embedding images for 1 glyphs for this strike.
Strike ppem set to 109.
width 136
height 128
ascent 1900
descent 500
upem 2048
y_ppem 109
x_bearing 0
line_height 127.734375
line_ascent 101.123046875
y_bearing 101
```

Scaling up can easily put content outside bbox because we don't compute it currently (#238)

### DIFF
--- a/src/nanoemoji/color_glyph.py
+++ b/src/nanoemoji/color_glyph.py
@@ -49,28 +49,40 @@ def _shift_origin_0_0(
     return (-view_box.x * x_scale, -view_box.y * y_scale)
 
 
-def map_viewbox_to_font_emsquare(view_box: Rect, upem: int) -> Affine2D:
+def map_viewbox_to_font_emsquare(
+    view_box: Rect, upem: int, user_transform: Affine2D
+) -> Affine2D:
     x_scale, y_scale = _scale_viewbox_to_emsquare(view_box, upem)
     # flip y axis
     y_scale = -y_scale
     # shift so things are in the right place
     dx, dy = _shift_origin_0_0(view_box, x_scale, y_scale)
     dy = dy + upem
-    return Affine2D(x_scale, 0, 0, y_scale, dx, dy)
+    affine = Affine2D(x_scale, 0, 0, y_scale, dx, dy)
+    return Affine2D.compose_ltr((affine, user_transform))
 
 
 # https://docs.microsoft.com/en-us/typography/opentype/spec/svg#coordinate-systems-and-glyph-metrics
-def map_viewbox_to_otsvg_emsquare(view_box: Rect, upem: int) -> Affine2D:
+def map_viewbox_to_otsvg_emsquare(
+    view_box: Rect, upem: int, user_transform: Affine2D
+) -> Affine2D:
     x_scale, y_scale = _scale_viewbox_to_emsquare(view_box, upem)
     dx, dy = _shift_origin_0_0(view_box, x_scale, y_scale)
 
     # shift so things are in the right place
     dy = dy - upem
-    return Affine2D(x_scale, 0, 0, y_scale, dx, dy)
+    affine = Affine2D(x_scale, 0, 0, y_scale, dx, dy)
+    return Affine2D.compose_ltr((affine, user_transform))
 
 
-def _get_gradient_transform(grad_el, shape_bbox, view_box, upem) -> Affine2D:
-    transform = map_viewbox_to_font_emsquare(view_box, upem)
+def _get_gradient_transform(
+    upem: int,
+    user_transform: Affine2D,
+    grad_el: etree.Element,
+    shape_bbox: Rect,
+    view_box: Rect,
+) -> Affine2D:
+    transform = map_viewbox_to_font_emsquare(view_box, upem, user_transform)
 
     gradient_units = grad_el.attrib.get("gradientUnits", "objectBoundingBox")
     if gradient_units == "objectBoundingBox":
@@ -85,7 +97,13 @@ def _get_gradient_transform(grad_el, shape_bbox, view_box, upem) -> Affine2D:
     return transform
 
 
-def _parse_linear_gradient(grad_el, shape_bbox, view_box, upem, shape_opacity=1.0):
+def _parse_linear_gradient(
+    config: FontConfig,
+    grad_el: etree.Element,
+    shape_bbox: Rect,
+    view_box: Rect,
+    shape_opacity: float = 1.0,
+):
     gradient = SVGLinearGradient.from_element(grad_el, view_box)
 
     p0 = Point(gradient.x1, gradient.y1)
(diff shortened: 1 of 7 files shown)


src/nanoemoji/config.py
```diff
@@ -203,6 +214,12 @@ def load(config_file: Path = None, additional_srcs: Tuple[Path] = None) -> FontC
     output_file = _pop_flag(config, "output_file")
     color_format = _pop_flag(config, "color_format")
     upem = int(_pop_flag(config, "upem"))
+    ascent = int(_pop_flag(config, "ascent"))
+    descent = int(_pop_flag(config, "descent"))
+    transform = _pop_flag(config, "transform")
+    if not isinstance(transform, Affine2D):
```
anthrotype, line 220: isn't this flag value always going to be a string? the `if` may be unnecessary

src/nanoemoji/color_glyph.py
```diff
@@ -328,8 +330,18 @@ def create(
                     font_config.ignore_reuse_error,
                 )
             )
+        if user_transform is None:
+            user_transform = Affine2D.identity()
+        print("user_transform", user_transform)
```
anthrotype, line 335: all the print() will have to be replaced with equivalent logging.info() or debug()

anthrotype (approved):

src/nanoemoji/write_font.py
```diff
@@ -138,22 +138,24 @@ class ColorGenerator(NamedTuple):
...
+    ufo.info.ascender = config.ascent
+    ufo.info.descender = config.descent
+    ufo.info.unitsPerEm = config.upem
 
     # Must have .notdef and Win 10 Chrome likes a blank gid1 so make gid1 space
     ufo.newGlyph(".notdef")
     space = ufo.newGlyph(".space")
     space.unicodes = [0x0020]
-    space.width = upem
+    space.width = config.upem
```
anthrotype, line 154: should you also scale the default width by the user-provided transform? otherwise it won't be centered anymore.

Also, maybe the width itself could be a CLI/config option since it is an interesting metric like ascent/descent

anthrotype (commented):
The user transform should also affect the way we map the linear/radial gradient coordinates to font space. Here you are only transforming the SVG paths, but not also the gradient coordinates.

src/nanoemoji/write_font.py
rsheeter: +1 to width config; will add.

rsheeter: If the width is a config option I'm less sure it should be subject to transform; thoughts?

src/nanoemoji/config.py
rsheeter: the default Config.transform is Affine2D.identity

src/nanoemoji/write_font.py
anthrotype: don't transform if provided explicitly. the user can do the math

src/nanoemoji/color_glyph.py
rsheeter: shouldn't be any when done, I just found them helpful to see when testing a couple of glyphs

src/nanoemoji/write_font.py
rsheeter: agreed and so implemented

rsheeter:
Updated to impact gradients as well

rsheeter:
For example, `nanoemoji ~/oss/noto-emoji/svg/emoji_u1f977.svg --transform "scale(1.17) translate(0, -200)" --width 1275 --descent -250` looks fairly close to Noto CBDT

src/nanoemoji/color_glyph.py
```diff
@@ -211,15 +237,15 @@ def shape_cache_key(self):
         return (self.path, self.reuses)
 
 
-def _paint(debug_hint, upem, picosvg, shape):
+def _paint(debug_hint: str, config: FontConfig, picosvg: SVG, shape: SVGPath):
```
anthrotype, line 240: optional: annotate the return value as well (-> Paint), since you annotated all the parameters

src/nanoemoji/write_font.py
```diff
@@ -267,6 +269,11 @@ def _draw_glyph_extents(ufo: ufoLib2.Font, glyph: Glyph):
     pen = glyph.getPen()
     pen.moveTo((0, 0))
     pen.lineTo((ufo.info.unitsPerEm, ufo.info.unitsPerEm))
+
+    # TEMPORARY; uncomment to draw outside basic extents. @anthrotype has a PR in flight to fix properly.
```
anthrotype, line 273: let's merge your PR first, then I'll update and rebase mine

src/nanoemoji/write_font.py
```diff
@@ -440,19 +447,12 @@ def _ensure_codepoints_will_have_glyphs(ufo, glyph_inputs):
 
 def _generate_color_font(config: FontConfig, inputs: Iterable[InputGlyph]):
     """Make a UFO and optionally a TTFont from svgs."""
-    ufo = _ufo(config.family, config.upem, config.keep_glyph_names)
+    print("_generate_color_font", config.transform)
```
anthrotype, line 450: remove stray print() or turn into logging.debug

anthrotype (approved):

anthrotype:
I rebased #252, let's continue there

yisibl:
So, we can use `--transform=translate(0, -200)` to adjust the position of the icon on the baseline, right?

PR #251, opened 2021-03-04
Add font version_major and version_minor options

yisibl:
https://docs.microsoft.com/en-us/typography/opentype/otspec181/name#name-ids:~:text=Version%20string

> Version string. Should begin with the syntax ‘Version <number>.<number>’ (upper case, lower case, or mixed, with a space between “Version” and the number).
>
>The string must contain a version number of the following form: one or more digits (0-9) of value less than 65,535, followed by a period, followed by one or more digits of value less than 65,535. Any character other than a digit will terminate the minor number. A character such as “;” is helpful to separate different pieces of version information.
>
>The first such match in the string can be used by installation software to compare font versions. Note that some installers may require the string to start with “Version ”, followed by a version number as above.

### DIFF
--- a/src/nanoemoji/config.py
+++ b/src/nanoemoji/config.py
@@ -49,13 +49,14 @@ _COLOR_FORMATS = [
 
 # we use None as a sentinel for flag not set; FontConfig class has the actual defaults.
 # CLI flags override config file (which overrides default FontConfig).
-flags.DEFINE_string("config", None, "Config file")
+flags.DEFINE_string("config", None, "Config file.")
 flags.DEFINE_integer("upem", None, "Units per em.")
 flags.DEFINE_integer("width", None, "Width.")
 flags.DEFINE_integer("ascent", None, "Ascender")
 flags.DEFINE_integer("descent", None, "Descender.")
 flags.DEFINE_string("transform", None, "User transform, in font coordinates.")
-
+flags.DEFINE_integer("version_major", None, "Major version.")
+flags.DEFINE_integer("version_minor", None, "Minor version.")
 flags.DEFINE_string("family", None, "Family name.")
 flags.DEFINE_string("output_file", None, "Output filename.")
 flags.DEFINE_enum(
@@ -65,12 +66,12 @@ flags.DEFINE_enum(
     "Type of font to generate.",
 )
 flags.DEFINE_enum(
-    "output", None, ["ufo", "font"], "Whether to output a font binary or a UFO"
+    "output", None, ["ufo", "font"], "Whether to output a font binary or a UFO."
 )
 flags.DEFINE_bool(
     "keep_glyph_names", None, "Whether or not to store glyph names in the font."
 )
-flags.DEFINE_bool("clip_to_viewbox", None, "Whether to clip content outside viewbox")
+flags.DEFINE_bool("clip_to_viewbox", None, "Whether to clip content outside viewbox.")
 flags.DEFINE_float(
     "reuse_tolerance",
     None,
@@ -115,6 +116,8 @@ class FontConfig(NamedTuple):
     ascent: int = 950  # default based on Noto Emoji
     descent: int = -250  # default based on Noto Emoji
     transform: Affine2D = Affine2D.identity()
+    version_major: int = 1
+    version_minor: int = 0
     reuse_tolerance: float = 0.1
     ignore_reuse_error: bool = True
     keep_glyph_names: bool = False
@@ -145,6 +148,8 @@ def write(dest: Path, config: FontConfig):
         "ascent": config.ascent,
         "descent": config.descent,
         "transform": config.transform.tostring(),
+        "version_major": config.version_major,
+        "version_minor": config.version_minor,
         "reuse_tolerance": config.reuse_tolerance,
         "ignore_reuse_error": config.ignore_reuse_error,
         "keep_glyph_names": config.keep_glyph_names,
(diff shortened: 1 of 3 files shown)


anthrotype (commented):
this looks good. Could you add a test as well?

tests/write_font_test.py
```diff
@@ -41,6 +41,34 @@ def test_keep_glyph_names(svgs, color_format, keep_glyph_names):
         assert ttfont["post"].formatType == 3.0
         assert ufo.glyphOrder != ttfont.getGlyphOrder()
 
+@pytest.mark.parametrize(
+    "svgs", [("rect.svg", "one-o-clock.svg")]
+)
+@pytest.mark.parametrize(
+    "color_format", ["glyf", "cff2_colr_0", "cff2_colr_1", "cff_colr_0", "cff_colr_1", "glyf_colr_0", "glyf_colr_1", "picosvg", "untouchedsvg"]
```
yisibl, line 48: Do we need to test all font formats?

anthrotype: i don't think so, the version setting is orthogonal to the color font format and applies to all of them indistinctly as they all will have a head table and a name table, like any opentype fonts. I think you can just test a couple, like glyf_colr_1 and cff_colr_1

yisibl: Done, PTAL.

src/nanoemoji/config.py
```diff
@@ -47,8 +47,10 @@
 
 # we use None as a sentinel for flag not set; FontConfig class has the actual defaults.
 # CLI flags override config file (which overrides default FontConfig).
-flags.DEFINE_string("config", None, "Config file")
+flags.DEFINE_string("config", None, "Config file.")
 flags.DEFINE_integer("upem", None, "Units per em.")
+flags.DEFINE_integer("version_major", None, "Major version.")
```
anthrotype, line 52:
```suggestion
flags.DEFINE_integer("version_major", 1, "Major version.")
```

src/nanoemoji/config.py
```diff
@@ -47,8 +47,10 @@
...
+flags.DEFINE_integer("version_minor", None, "Minor version.")
```
anthrotype, line 53:
```suggestion
flags.DEFINE_integer("version_minor", 0, "Minor version.")
```

tests/test_helper.py
```diff
@@ -59,6 +59,7 @@ def color_font_config(config_overrides, svgs, tmp_dir=None):
             upem=100,
             keep_glyph_names=True,
             fea_file=fea_file,
+            # version_minor=2,
```
anthrotype, line 62: what's this?

tests/write_font_test.py
```diff
@@ -42,6 +42,40 @@ def test_keep_glyph_names(svgs, color_format, keep_glyph_names):
...
+        "cff_colr_1",
+        "glyf_colr_0",
+        "glyf_colr_1",
+        "picosvg",
+    ],
+)
+@pytest.mark.parametrize("version_major", [12, 9])
+@pytest.mark.parametrize("version_minor", [0, "008"])
+# Test version 16.28
+def test_version_minor(svgs, color_format, version_major, version_minor):
```
anthrotype, line 59: why do you parametrize the `version_major ` and `version_minor` if you then discard the parametrized value and hard-code 16.28 in the body of the test function?

tests/write_font_test.py
```diff
@@ -42,6 +42,40 @@ def test_keep_glyph_names(svgs, color_format, keep_glyph_names):
...
+            "version_minor": 28,
+        },
+        svgs,
+    )
+    ufo, ttfont = write_font._generate_color_font(config, glyph_inputs)
+    ttfont = test_helper.reload_font(ttfont)
+
+    if version_major:
+        assert ufo.info.versionMajor == 16
+    elif version_minor:
```
anthrotype, line 73: this `if ... elif ...` block does not make any sense. why don't just assert the value is as expected, why if?

tests/write_font_test.py
```diff
@@ -42,6 +42,40 @@ def test_keep_glyph_names(svgs, color_format, keep_glyph_names):
...
+        assert ufo.info.versionMinor == 28
+    # else:
+    # TODO: test default version 1.000
```
anthrotype, line 76: you may as well add a separate test for that

anthrotype (commented):
I left some comments.

In the tests you want to check that the version string (nameID=5) in the name table is as expected. E.g.

```python
assert "Version 16.28" in ttfont["name"].getDebugName(5)
```

src/nanoemoji/config.py
anthrotype: Actually, it's correct like this. The flags default should be all `None`, in this way the FontConfig default values will be used when a config parameter is not overridden in a config file nor with a CLI flag. Am I right @rsheeter?

rsheeter: Correct, we take None as "no flag provided"; the default goes on the Config struct

tests/test_helper.py
yisibl: Removed.

tests/write_font_test.py
yisibl: Done

tests/write_font_test.py
yisibl: Done

tests/write_font_test.py
yisibl: Add `test_default_version`

yisibl:
> I left some comments.
> 
> In the tests you want to check that the version string (nameID=5) in the name table is as expected. E.g.
> 
> ```python
> assert "Version 16.28" in ttfont["name"].getDebugName(5)
> ```

Done.

Add openTypeNameVersion and openTypeNameUniqueID test case

anthrotype:
sorry, I merged before you finished, I thought you were done. Anyway please have a look at https://github.com/googlefonts/nanoemoji/pull/253

PR #252, opened 2021-03-05
Compute bbox correctly (for statics)

anthrotype:
Helps with https://github.com/googlefonts/nanoemoji/issues/238

### DIFF
--- a/src/nanoemoji/write_font.py
+++ b/src/nanoemoji/write_font.py
@@ -454,7 +454,6 @@ def _ensure_codepoints_will_have_glyphs(ufo, glyph_inputs):
 
 def _generate_color_font(config: FontConfig, inputs: Iterable[InputGlyph]):
     """Make a UFO and optionally a TTFont from svgs."""
-    print("_generate_color_font", config.transform)
     ufo = _ufo(config)
     _ensure_codepoints_will_have_glyphs(ufo, inputs)
     base_gid = len(ufo.glyphOrder)
--- a/src/nanoemoji/write_font.py
+++ b/src/nanoemoji/write_font.py
@@ -21,7 +21,7 @@ from absl import logging
 from collections import defaultdict
 import csv
 from fontTools import ttLib
-from fontTools.misc.arrayTools import unionRect
+from fontTools.misc.arrayTools import rectArea, unionRect
 from fontTools.ttLib.tables import otTables as ot
 from itertools import chain
 from lxml import etree  # pytype: disable=import-error
@@ -262,17 +262,16 @@ def _create_glyph(color_glyph: ColorGlyph, painted_layer: PaintedLayer) -> Glyph
 
 
 def _draw_glyph_extents(
-    ufo: ufoLib2.Font, glyph: Glyph, bounds: Optional[Tuple[float, float, float, float]]
+    ufo: ufoLib2.Font, glyph: Glyph, bounds: Tuple[float, float, float, float]
 ):
     # apparently on Mac (but not Linux) Chrome and Firefox end up relying on the
     # extents of the base layer to determine where the glyph might paint. If you
     # leave the base blank the COLR glyph never renders.
 
-    start = (0, 0)
-    end = (ufo.info.unitsPerEm, ufo.info.unitsPerEm)
-    if bounds is not None:
-        start = bounds[:2]
-        end = bounds[2:]
+    if rectArea(bounds) == 0:
+        return
+
+    start, end = bounds[:2], bounds[2:]
 
     pen = glyph.getPen()
     pen.moveTo(start)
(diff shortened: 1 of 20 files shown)


src/nanoemoji/write_font.py
```diff
@@ -356,6 +361,13 @@ def _ufo_colr_layers(colr_version, colors, color_glyph, glyph_cache):
         else:
             glyph = glyph_cache[glyph_cache_key]
 
+        glyph_bbox = glyph.getControlBounds(color_glyph.ufo)
```
rsheeter, line 364: Optional, alternately compute bounds on ColorGlyph alongside of computing the painted layers so here you just access, e.g. color_glyph.bbox

src/nanoemoji/write_font.py
```diff
@@ -257,16 +259,22 @@ def _create_glyph(color_glyph: ColorGlyph, painted_layer: PaintedLayer) -> Glyph
     return glyph
 
 
-def _draw_glyph_extents(ufo: ufoLib2.Font, glyph: Glyph):
+def _draw_glyph_extents(
+    ufo: ufoLib2.Font, glyph: Glyph, bounds: Optional[Tuple[float, float, float, float]]
```
rsheeter, line 263: bounds seems like it should be a non-optional Rect?

src/nanoemoji/write_font.py
```diff
@@ -257,16 +259,22 @@ def _create_glyph(color_glyph: ColorGlyph, painted_layer: PaintedLayer) -> Glyph
...
+):
     # apparently on Mac (but not Linux) Chrome and Firefox end up relying on the
     # extents of the base layer to determine where the glyph might paint. If you
     # leave the base blank the COLR glyph never renders.
 
-    # TODO we could narrow this to bbox to cover all layers
+    start = (0, 0)
+    end = (ufo.info.unitsPerEm, ufo.info.unitsPerEm)
```
rsheeter, line 270: We should have bounds unless nothing at all is drawn, correct? If so, I think just not drawing anything in the base glyph would make sense, e.g. if bounds.area == 0: exit

src/nanoemoji/write_font.py
rsheeter: Not quite sure what glyph.getControlBounds does ... how is this accounting for paint me that glyph with that transform?

rsheeter:
Fantastic to have.

Can you add a test that demonstrates that we correctly calculate teh bbox in the presence of transformation? E.g. rotate a rect with a large width and small height by 90 around some center and show the bbox is for a small width/large height? - currently wasn't quite clear to me that the DAG effects were factored in.

I suggest adding a dedicated bbox test; I think it deserves it.

src/nanoemoji/write_font.py
anthrotype: We can't compute it on the painted layers, those still contain svg paths. We have to compute the bounds on the layer glyph coordinates after they have been mapped to UPEM space. It's also easier to do it on a UFO glyph with the getControlBounds method than on the svg paths. the `ufoLib2.Glyph.getControlBounds` method uses the fonttools ControlBoundsPen to return the bounding box for the control points of that glyph. It's `None` when the glyph is empty and has no contours nor components. The reason you need to pass the glyphSet parameter to getControlBounds is to be able to resolve component references, as the glyph object does not have a reference to the layer it is in (I'm passing the ufo object itself as the glyphSet for resolving component references, that's equivalent of passing the UFO default layer)

src/nanoemoji/write_font.py
anthrotype: made it non-optional

src/nanoemoji/write_font.py
anthrotype: made it to return early if rectangle area is 0 (unlikely but you never know)

src/nanoemoji/write_font.py
rsheeter: > We have to compute the bounds on the layer glyph coordinates after they have been mapped to UPEM space and user-transformed

Or compute the bbox on the layers then map it into upem space?

> return the bounding box for the control points of that glyph

How are we accounting for the DAG's impact on the glyph bounds? For example, if my graph says PaintScaled(double size, center 100,100) > PaintGlyph(a box from 100,100 to 200,200) I would expect glyph control bounds to be 100,100 => 200,200 (not accounting for the scale).

anthrotype: > How are we accounting for the DAG's impact on the glyph bounds? 

damn. We aren't. Need rethink

anthrotype: > compute the bbox on the layers then map it into upem space?

i found it much easier to reuse ready made method for computing bbox of UFO glyphs, rather than work on svg paths

anthrotype: > if my graph says PaintScaled...

FWIW note that right now we aren't using any of that

rsheeter (approved):
LG but per IM let's not mark as fixing #238; we should hold that until we support arbitrary graphs.

rsheeter:
Updated comment so it won't close the issue, merging

PR #253, opened 2021-03-08
test name table contains version string, simplify version tests

anthrotype:
@yisibl PTAL

### DIFF
--- a/tests/write_font_test.py
+++ b/tests/write_font_test.py
@@ -43,60 +43,46 @@ def test_keep_glyph_names(svgs, color_format, keep_glyph_names):
         assert ufo.glyphOrder != ttfont.getGlyphOrder()
 
 
-@pytest.mark.parametrize("svgs", [("rect.svg", "one-o-clock.svg")])
 @pytest.mark.parametrize(
     "color_format",
     [
+        "glyf",
         "cff_colr_0",
-        "cff_colr_1",
-        "glyf_colr_0",
         "glyf_colr_1",
         "picosvg",
     ],
 )
-# Default version is 1.000
-def test_default_version(svgs, color_format):
-    config, glyph_inputs = test_helper.color_font_config(
-        {
-            "color_format": color_format,
-        },
-        svgs,
-    )
-    ufo, ttfont = write_font._generate_color_font(config, glyph_inputs)
-    ttfont = test_helper.reload_font(ttfont)
-
-    assert ufo.info.versionMajor == 1
-    assert ufo.info.versionMinor == 0
-
-
-@pytest.mark.parametrize("svgs", [("rect.svg", "one-o-clock.svg")])
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


tests/write_font_test.py
```diff
@@ -43,60 +43,46 @@ def test_keep_glyph_names(svgs, color_format, keep_glyph_names):
...
-        "cff_colr_0",
-        "cff_colr_1",
-        "glyf_colr_0",
-        "glyf_colr_1",
-        "picosvg",
+        (None, None, "1.000"),  # default
+        (1, 2, "1.002"),
+        (None, 1, "1.001"),
+        (2, None, "2.000"),
+        (16, 28, "16.028"),
```
anthrotype, line 62: @yisibl note how the version_minor gets left-padded with zeros, that's how OpenType versions are supposed to work. If you meant `16.280` you have to set version minor to `280` otherwise you get `16.028`. Does that make sense?

tests/write_font_test.py
```diff
@@ -43,60 +43,46 @@ def test_keep_glyph_names(svgs, color_format, keep_glyph_names):
...
+        config_overrides, ("rect.svg", "one-o-clock.svg")
     )
     ufo, ttfont = write_font._generate_color_font(config, glyph_inputs)
     ttfont = test_helper.reload_font(ttfont)
 
-    assert ufo.info.versionMajor == 16
-    assert ufo.info.versionMinor == 28
+    assert ufo.info.versionMajor == version_major
+    assert ufo.info.versionMinor == version_minor
+    assert ttfont["name"].getDebugName(nameID=5).startswith(f"Version {expected}")
```
yisibl, line 85: Is this more intuitive?

`assert ttfont["name"].getDebugName(nameID=5) == ("Version {expected}")`

yisibl: Consider adding tests for openTypeNameUniqueID? 

I haven't had time to submit in my PR. https://github.com/googlefonts/nanoemoji/commit/5576797#diff-911b8272daeaab9d9d5652de24b083f29c6e16fa70b8784067247dbf479db167R76

tests/write_font_test.py
yisibl: I think we can consider adding `16.280` with three decimal places, and even a version number with four decimal places, which helps us understand what the final version number is.

Learned, very good example. Sorry, I just started to learn to use pytest.

tests/write_font_test.py
anthrotype: I intentionally made it only check that the nameID version nameID starts with the expected string. The compiler may add extra stuff separated by semicolon which we are not interested in testing in here. Yes, we could also test TT unique id, or even head.fontRevision, but those are kind of secondary and dependent on the primary version string. I think this is enough.

yisibl: LGTM

Issue #230, opened 2021-01-30
The shape is stretched after the font is generated


yisibl:
```svg
<svg xmlns="http://www.w3.org/2000/svg" width="245" height="168" viewBox="0 0 245 168">
  <g fill="none">
    <path fill="#ED7A1F" d="M167.346044,84.327826 C167.346044,118.170061 146.959974,148.679972 115.693825,161.630835 C84.4276764,174.581698 48.4387826,167.423062 24.5087085,143.492988 C0.578634349,119.562914 -6.58000169,83.5740199 6.37086113,52.3078713 C19.321724,21.0417228 49.8316349,0.655652069 83.6738703,0.655652148 C105.921096,0.471358231 127.310687,9.22760131 143.042391,24.9593053 C158.774095,40.6910093 167.530338,62.0806002 167.346044,84.327826 Z"/>
    <path fill="#849EFD" d="M244.476575,94.9770119 C244.476575,135.215721 211.007706,167.695738 156.772924,167.695737 C130.218431,168.55446 105.305171,154.877976 91.7738251,132.013567 C78.2424788,109.149158 78.2424788,80.7288003 91.7738251,57.8643914 C105.305171,34.9999825 130.218431,21.3234979 156.772924,22.1822205 C211.083771,22.1822205 244.476575,54.7383027 244.476575,94.9770119 L244.476575,94.9770119 Z"/>
  </g>
</svg>
```

![image](https://user-images.githubusercontent.com/2784308/106354014-acde7080-6329-11eb-8fed-48305fe671d5.png)

![image](https://user-images.githubusercontent.com/2784308/106354029-cf708980-6329-11eb-8bbf-7c7aec54ae41.png)

rsheeter:
@anthrotype just as you foretold non-square viewboxes have struck (`viewBox="0 0 245 168"`) :)

@yisibl definitely a bug on our side that needs fixing. Until fixed you'll get best results if your viewbox is square.

yisibl:
That makes sense. In the technology of generating [iconfont](https://www.iconfont.cn/), I always scale the viewBox to 1024 * 1024。

anthrotype:
[...]

PR #255, opened 2021-03-09
scale height to (ascender-descender), preserve aspect, center width

anthrotype:
Fixes https://github.com/googlefonts/nanoemoji/issues/2#issuecomment-794250786 and https://github.com/googlefonts/nanoemoji/issues/230

Previously we were mapping the SVG viewBox to UPEM square (0, 0, upem, upem), stretching it when the aspect ratio was not square, and setting the advance width to the same monospaced value.

After this PR:

1) the viewbox height is scaled to match the sum of font (ascender - descender), and the width is scaled by the same factor (thus preserving the aspect ratio); the design centered horizontally within the glyph advance width. This is what NotoColorEmoji also does, and it makes the emoji designs look more visually centered when set alongside normal text (#2).
2) changed the default FontConfig.width to the same value of NotoColorEmoji: this is wider than default upem, and a bit wider than the default (ascender - descender). Rationale: if it looks good with noto-emoji, should look good enough for other emoji designs too
3) the advance width is now proportional to the aspect ratio: i.e. if viewbox width == height (square viewbox), then the nominal width doesn't change, and if all input SVGs have square ratio the font will be monospaced (all advance widths == config.width); if a viewbox is non-square (like the example duck in #230), then the advance width is also scaled to fit the scaled viewbox; thus a glyph advance width can be wider or narrower than the nominal width depending on the input SVG aspect ratio.

### DIFF
--- a/src/nanoemoji/color_glyph.py
+++ b/src/nanoemoji/color_glyph.py
@@ -37,52 +37,58 @@ from typing import Generator, NamedTuple, Optional, Sequence, Tuple
 import ufoLib2
 
 
-def _scale_viewbox_to_emsquare(view_box: Rect, upem: int) -> Tuple[float, float]:
-    # scale to font upem
-    return (upem / view_box.w, upem / view_box.h)
+def _scale_viewbox_to_font_metrics(
+    view_box: Rect, ascender: int, descender: int, width: int
+):
+    # scale height to (ascender - descender)
+    scale = (ascender - descender) / view_box.h
+    # shift so width is centered
+    dx = (width - scale * view_box.w) / 2
+    return Affine2D.compose_ltr(
+        (
+            # first normalize viewbox origin
+            Affine2D(1, 0, 0, 1, -view_box.x, -view_box.y),
+            Affine2D(scale, 0, 0, scale, dx, 0),
+        )
+    )
 
 
-def _shift_origin_0_0(
-    view_box: Rect, x_scale: float, y_scale: float
-) -> Tuple[float, float]:
-    # shift so origin is 0,0
-    return (-view_box.x * x_scale, -view_box.y * y_scale)
-
-
-def map_viewbox_to_font_emsquare(
-    view_box: Rect, upem: int, user_transform: Affine2D
+def map_viewbox_to_font_space(
+    view_box: Rect, ascender: int, descender: int, width: int, user_transform: Affine2D
 ) -> Affine2D:
-    x_scale, y_scale = _scale_viewbox_to_emsquare(view_box, upem)
-    # flip y axis
-    y_scale = -y_scale
-    # shift so things are in the right place
-    dx, dy = _shift_origin_0_0(view_box, x_scale, y_scale)
-    dy = dy + upem
-    affine = Affine2D(x_scale, 0, 0, y_scale, dx, dy)
-    return Affine2D.compose_ltr((affine, user_transform))
+    return Affine2D.compose_ltr(
+        [
+            _scale_viewbox_to_font_metrics(view_box, ascender, descender, width),
+            # flip y axis and shift so things are in the right place
+            Affine2D(1, 0, 0, -1, 0, ascender),
+            user_transform,
+        ]
+    )
 
 
 # https://docs.microsoft.com/en-us/typography/opentype/spec/svg#coordinate-systems-and-glyph-metrics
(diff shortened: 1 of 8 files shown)


src/nanoemoji/color_glyph.py
```diff
@@ -37,52 +37,58 @@
...
-def _shift_origin_0_0(
-    view_box: Rect, x_scale: float, y_scale: float
-) -> Tuple[float, float]:
-    # shift so origin is 0,0
-    return (-view_box.x * x_scale, -view_box.y * y_scale)
+def _scale_viewbox_to_font_metrics(
+    view_box: Rect, ascender: int, descender: int, width: int
+):
+    # scale height to (ascender - descender)
+    scale = (ascender - descender) / view_box.h
```
rsheeter, line 44: suggest assert descender <= 0; conveniently makes it much more obvious that we are *adding* descender to ascender here.

src/nanoemoji/color_glyph.py
```diff
@@ -373,16 +385,24 @@ def transform_for_font_space(self):
...
             return Affine2D.identity()
-        return map_viewbox_to_otsvg_emsquare(
-            self.svg.view_box(), self.ufo.info.unitsPerEm, self.user_transform
+        return map_viewbox_to_otsvg_space(
+            self.svg.view_box(),
+            self.ufo.info.ascender,
+            self.ufo.info.descender,
+            self.ufo[self.glyph_name].width,
+            self.user_transform,
         )
```
rsheeter, line 406: Optional, consider a common fn akin to:

```python
def _transform(self, map_fn):
        if not self._has_viewbox_for_transform():
            return Affine2D.identity()
        return map_fn(
            self.svg.view_box(),
            self.ufo.info.ascender,
            self.ufo.info.descender,
            self.ufo[self.glyph_name].width,
            self.user_transform,
        )

def transform_for_otsvg_space(self):
  return self._transform(map_viewbox_to_otsvg_space)

def transform_for_font_space(self):
  return self._transform(map_viewbox_to_font_space)
```

tests/color_glyph_test.py
```diff
@@ -41,35 +43,81 @@ def _nsvg(filename):
...
+        (
+            "0 0 1024 1024",
+            1024,
+            1275,
+            950,
+            -250,
+            Affine2D(1.171875, 0, 0, -1.171875, 37.5, 950),
+            1275,
+        ),
+        # viewbox wider than tall
```
rsheeter, line 82: suggest making these more descriptive, e.g. viewbox wider than tall so <what transform is expected>

rsheeter (approved):

yisibl:
A great improvement, with PR https://github.com/googlefonts/nanoemoji/pull/256, the generated color icons have almost the same effect as the black and white icons in iconfont.

image

rsheeter:
ty for sharing that example!

drott:
Yeyh for fixing #2, big thanks!

yisibl:
@drott Next is waiting for you to fix Chrome's COLR font blurring problem.

drott:
> @drott Next is waiting for you to fix Chrome's COLR font blurring problem.

Let's stay on topic on this PR. If you are referring to https://bugs.chromium.org/p/chromium/issues/detail?id=1112051 - please comment (and optionally attach screenshots there) if you observe this issue for COLRv1 as well.

anthrotype:
Two new notable changes:

1) there was a bug in that I was not taking into account the particular glyph advance width when mapping gradient coordinates to font space, but was always assuming the default config.width (I fixed it with 09c8af0, test 20e45db).

2) I changed the logic for computing the glyph advance width given the font height (ascender-descender) and the viewbox aspect ratio. We now use that only when it exceeds the default advance width. If smaller (viewbox taller than wider), we use the default width and place things centered horizontally. The way I set it up initially did not make sense in all cases, because simply scaling the default width by the aspect ratio would produce an outsized effect on that extra left/right padding, which I believe is not what one really wants. The logic now is this:

```python
def _color_glyph_advance_width(view_box: Rect, config: FontConfig) -> int:
    # Scale advance width proportionally to viewbox aspect ratio.
    # Use the default advance width if it's larger than the proportional one.
    font_height = config.ascent - config.descent  # descent <= 0
    return max(config.width, round(font_height * view_box.w / view_box.h))
```

rsheeter:
LGTM

PR #256, opened 2021-03-10
Set openTypeOS2TypoLineGap=0

yisibl:
[ufo2ft](https://github.com/googlefonts/ufo2ft/blob/97eb6734397dcb2bc395775c4e342d74cd6474a3/Lib/ufo2ft/fontInfoData.py#L229-L238) currently calculated value is not equal to 0

Mainly refer to the conclusion of this article:
https://simoncozens.github.io/fonts-and-layout/opentype.html#vertical-metrics-hhea-and-os2:~:text=recommended%20method

> To get you started, here is our recommended method (distilled from a [discussion on Typedrawers](http://typedrawers.com/discussion/1705)):
> 
> - The sTypoAscender minus the sTypoDescender should equal the unit square. (Usually 1000 units.)
> 
> - sTypoLinegap should be the default linespacing distance for your intended language environment.
> 
> - lineGap should be zero.
> 
> - usWinAscent and usWinDescent should be set so that no clipping occurs. If your font contains glyphs with tall, stacked accents - for instance, the Vietnamese letter LATIN CAPITAL LETTER A WITH BREVE AND HOOK ABOVE (Ẳ) - you will need to ensure that these values can accommodate the highest (and lowest) possible values of your shaped text. They should also be set so that they sum to at least the value of sTypoAscender - sTypoDescender + sTypoLinegap.
> 
> - ascent and descent should be set to the same values as usWinAscent and usWinDescent, remembering that usWinDescent is positive and descent is negative.
> 
> - Bit 7 of fsSelection should be turned on.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,11 +47,9 @@ setup(
         ],
     },
     python_requires=">=3.6",
-
     # this is for type checker to use our inline type hints:
     # https://www.python.org/dev/peps/pep-0561/#id18
     package_data={"picosvg": ["py.typed"]},
-
     # metadata to display on PyPI
     author="Rod S",
     author_email="rsheeter@google.com",
--- a/src/nanoemoji/write_font.py
+++ b/src/nanoemoji/write_font.py
@@ -152,6 +152,11 @@ def _ufo(config: FontConfig) -> ufoLib2.Font:
     ufo.info.versionMajor = config.version_major
     ufo.info.versionMinor = config.version_minor
 
+    # Set OS/2 sTypoLineGap, make it equal to hhea lineGap
+    # https://docs.microsoft.com/en-us/typography/opentype/spec/os2#stypolinegap
+    # The default line gap is OS/2.sTypoAscender - OS/2.sTypoDescender + OS/2.sTypoLineGap
+    ufo.info.openTypeOS2TypoLineGap = 0
+
     # Must have .notdef and Win 10 Chrome likes a blank gid1 so make gid1 space
     ufo.newGlyph(".notdef")
     space = ufo.newGlyph(".space")
--- a/tests/write_font_test.py
+++ b/tests/write_font_test.py
@@ -84,6 +84,28 @@ def test_version(color_format, version_major, version_minor, expected):
     assert ttfont["name"].getDebugName(nameID=5).startswith(f"Version {expected}")
 
 
+@pytest.mark.parametrize(
+    "color_format",
+    [
+        "glyf",
+        "cff_colr_0",
+        "glyf_colr_1",
+        "picosvg",
+    ],
+)
+def test_default_line_gap(color_format):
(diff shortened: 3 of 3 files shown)


src/nanoemoji/config.py
```diff
@@ -113,6 +113,9 @@ class FontConfig(NamedTuple):
     color_format: str = "glyf_colr_1"
     upem: int = 1024
     width: int = 1024
+    # TODO: How OpenType Works says:
+    # The sTypoAscender minus the sTypoDescender should equal the unit square.
+    # Do we need to adjust the default value here?
     ascent: int = 950  # default based on Noto Emoji
     descent: int = -250  # default based on Noto Emoji
```
yisibl, line 120: @rsheeter PTAL

rsheeter: @anthrotype thoughts? - I'll note Noto does not appear to follow this rule

anthrotype:
In general, I agree that we need to provide a way for the user to customise the line gap, which is important when determining the overall line spacing, along with ascender and descender.

What happens now is that we are not setting the line gap explicitly in the UFO, so our font compiler (ufo2ft) uses its own fallback value --- `UPEM * 1.2 - ascender + descender, or 0 if negative` -- which may not be what the user actually wants.

I agree that openTypeOS2TypoLineGap=0 is a better default value than the current (implicit) one. We should also add a new `--linegap` CLI flag/config option to set it to something else.

src/nanoemoji/config.py
anthrotype: > sTypoAscender minus the sTypoDescender should equal the unit square

not true, that recommendation is bogus. One can set them to whatever makes sense for the design at hand.

anthrotype:
@yisibl if you can hold on for a day or two, I'd like to take care of this myself. But thanks for all the precious issue-filings!

src/nanoemoji/config.py
yisibl: It does not need to be equal the unit square.

I found some introductions about [UPM dogma](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#stypoascender). And the description of the [opentype spec](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#stypoascender): 
> t is not a general requirement that sTypoAscender - sTypoDescender be equal to unitsPerEm. These values should be set to provide default line spacing appropriate for the primary languages the font is designed to support.

yisibl:
@anthrotype Do you have any other details to consider? If just add a flag, I can try to do it. I am waiting for the correct lineGap to generate iconfont color fonts.

src/nanoemoji/config.py
yisibl: Comment removed.

anthrotype:
I'm working on this, please wait thank you.

yisibl:
Thank you.

PR #258, opened 2021-03-12
set default linegap to 0, allow to customise via --linegap option

anthrotype:
font vertical metrics is a contentious topic, here I want to keep things simple without opening the can of worms.

With this PR, we set the same default linegap that NotoColorEmoji.ttf uses, ie. 0 (which maximises compatibility across environments).
We set all ascenders/descenders to the same nominal values provided by the user config/CLI flags.
We set fsSelection bit 7 to force the use of typo-metrics for line spacing over the legacy OS/2 winAscent/winDescent.

Note that the winAscent/winDescent may still be relevant for defining clipping zones. However at this time, we don't set them explicitly to the actual font's yMin/yMax but use ufo2ft's fallback values (winAscent == ascender + linegap, winDescent == abs(descender)).
The assumption being emojis are usually contained well within the viewbox. We can revise that later if need be (or potentially fix that in ufo2ft itself).

### DIFF
--- a/src/nanoemoji/write_font.py
+++ b/src/nanoemoji/write_font.py
@@ -145,9 +145,21 @@ def _ufo(config: FontConfig) -> ufoLib2.Font:
     ufo.info.familyName = config.family
     # set various font metadata; see the full list of fontinfo attributes at
     # https://unifiedfontobject.org/versions/ufo3/fontinfo.plist/#generic-dimension-information
-    ufo.info.ascender = config.ascender
-    ufo.info.descender = config.descender
     ufo.info.unitsPerEm = config.upem
+    # we just use a simple scheme that makes all sets of vertical metrics the same;
+    # if one needs more fine-grained control they can fix up post build
+    ufo.info.ascender = (
+        ufo.info.openTypeHheaAscender
+    ) = ufo.info.openTypeOS2TypoAscender = config.ascender
+    ufo.info.descender = (
+        ufo.info.openTypeHheaDescender
+    ) = ufo.info.openTypeOS2TypoDescender = config.descender
+    ufo.info.openTypeHheaLineGap = ufo.info.openTypeOS2TypoLineGap = config.linegap
+    # set USE_TYPO_METRICS flag (OS/2.fsSelection bit 7) to make sure OS/2 Typo* metrics
+    # are preferred to define Windows line spacing over legacy WinAscent/WinDescent:
+    # https://docs.microsoft.com/en-us/typography/opentype/spec/os2#fsselection
+    ufo.info.openTypeOS2Selection = [7]
+
     # version
     ufo.info.versionMajor = config.version_major
     ufo.info.versionMinor = config.version_minor
--- a/tests/write_font_test.py
+++ b/tests/write_font_test.py
@@ -14,6 +14,7 @@
 
 
 from nanoemoji import write_font
+from nanoemoji.config import _DEFAULT_CONFIG
 from picosvg.svg_transform import Affine2D
 import pytest
 import test_helper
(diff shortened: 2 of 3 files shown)


src/nanoemoji/write_font.py
```diff
@@ -145,9 +145,21 @@ def _ufo(config: FontConfig) -> ufoLib2.Font:
...
+        ufo.info.openTypeHheaAscender
+    ) = ufo.info.openTypeOS2TypoAscender = config.ascender
+    ufo.info.descender = (
+        ufo.info.openTypeHheaDescender
+    ) = ufo.info.openTypeOS2TypoDescender = config.descender
+    ufo.info.openTypeHheaLineGap = ufo.info.openTypeOS2TypoLineGap = config.linegap
+    # set USE_TYPO_METRICS flag (OS/2.fsSelection bit 7) to make sure OS/2 Typo* metrics
+    # are preferred to define Windows line spacing over legacy WinAscent/WinDescent:
+    # https://docs.microsoft.com/en-us/typography/opentype/spec/os2#fsselection
+    ufo.info.openTypeOS2Selection = [7]
```
yisibl, line 161: For Font selection flags, do we want to enable Bit 6?

anthrotype: that one is special and is set automatically by the ufo2ft compiler depending on the style name (which is "Regular" by default). In any case, it's a separate issue.

src/nanoemoji/write_font.py
```diff
@@ -145,9 +145,21 @@ def _ufo(config: FontConfig) -> ufoLib2.Font:
...
     ufo.info.unitsPerEm = config.upem
+    # we just use a simple scheme that makes all sets of vertical metrics the same;
+    # if one needs more fine-grained control they can fix up post build
+    ufo.info.ascender = (
+        ufo.info.openTypeHheaAscender
+    ) = ufo.info.openTypeOS2TypoAscender = config.ascender
+    ufo.info.descender = (
+        ufo.info.openTypeHheaDescender
+    ) = ufo.info.openTypeOS2TypoDescender = config.descender
+    ufo.info.openTypeHheaLineGap = ufo.info.openTypeOS2TypoLineGap = config.linegap
```
yisibl, line 157: After comparing the ttx files, I found that sTypoLineGap is set to a non-zero value here. I am not sure if there will be problems in IE 11.
https://github.com/fontello/svg2ttf/commit/0f25b4081d91c209787bdfd999b2ef574461fedc

anthrotype: it is set to 0 by default. Just leave it as is if you want to keep it so

yisibl: The current behavior will set openTypeHheaLineGap to always be equal to openTypeOS2TypoLineGap, but the behavior of svg2ttf is that they are not equal. Please see this test case:

https://github.com/fontello/svg2ttf/pull/102/commits/66c503e185c7cb8786f4426fe47acdde68e70ddf#diff-a561630bb56b82342bc66697aee2ad96efddcbc9d150665abd6fb7ecb7c0ab2fR193-R199

anthrotype: Well, I argue that svg2ttf is wrong, and there's nothing bad with having hhea.lineGap be equal to OS/2.sTypoLineGap (in presence of USE_MY_METRICS flag). You know, that's exactly what I meant when I said that this topic is "contentious".
I _don't_ want to get sucked into this never ending discussions, so I kindly ask you to don't insist further. These defaults are good enough for supporting modern environments and they are correct for the limited scope of emoji fonts. If you need more fine-grained control, or you need to support some legacy environments, you can write a fontTools script to fix up the values after you have built the fonts.

yisibl: I just further confirm whether there is a problem with the implementation of svg2ttf. It is indeed possible not to consider IE11 too much at present. Thank you for your PR.

tests/write_font_test.py
```diff
@@ -84,6 +85,44 @@ def test_version(color_format, version_major, version_minor, expected):
...
+    hhea = ttfont["hhea"]
+    os2 = ttfont["OS/2"]
+
+    assert ufo.info.ascender == hhea.ascent == os2.sTypoAscender == ascender
+    assert ufo.info.descender == hhea.descent == os2.sTypoDescender == descender
+    assert hhea.lineGap == os2.sTypoLineGap == linegap
+    # check USE_TYPO_METRICS is set
+    assert os2.fsSelection & (1 << 7) != 0
+    # These are ufo2ft's fallback WinAscent/WinDescent, good enough for now.
+    # TODO: Set to the actual global yMin/yMax to prevent any clipping?
```
rsheeter, line 121: Should we file an issue for this?

rsheeter (approved):

tests/write_font_test.py
anthrotype: looks like there was one already: https://github.com/googlefonts/ufo2ft/issues/347

PR #259, opened 2021-03-12
use explicit --output_file instead of bash '>' redirection op

anthrotype:
On Windows, ninja uses the system API to create subprocesses that does not support the Bash `>` redirection operator so we must use an explicit `--output_file` command-line option for `nanoemoji.write_fea` and `nanoemoji.write_codepoints` sub-commands.

See https://github.com/googlefonts/nanoemoji/pull/240

NOTE: tests will fail, until similar picosvg PR is merged and released (and picosvg dependency bumped in nanoemoji): https://github.com/googlefonts/picosvg/pull/194

### DIFF
--- a/src/nanoemoji/write_fea.py
+++ b/src/nanoemoji/write_fea.py
@@ -21,6 +21,7 @@ from absl import app
 from absl import flags
 from nanoemoji import codepoints
 from nanoemoji import features
+from nanoemoji import util
 
 
 FLAGS = flags.FLAGS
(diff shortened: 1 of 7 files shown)


src/nanoemoji/write_fea.py
```diff
@@ -18,33 +18,25 @@
 
 
 from absl import app
+from absl import flags
 from nanoemoji import codepoints
-from nanoemoji.glyph import glyph_name
+from nanoemoji import features
```
anthrotype, line 23: I had to move `generate_fea` function to a separate `nanoemoji.features` submodule, otherwise defining a flag named `output_file` inside `nanoemoji.write_fea` (the CLI entry point) would create a duplicate flag conflict (as it's also defined in the main `nanoemoji.config` module; the `test_helpers.py` where importing from `nanoemoji.write_fea` and absl.flags was complaining). I think it's neater to keep the importable core logic separate from the modules that are meant to be executed from the console.

anthrotype:
merging so I can check if https://github.com/googlefonts/nanoemoji/pull/240 then pass on windows

Issue #266, opened 2021-04-09
Svg font does not support <use> in Safari


yisibl:
`nanoemoji --color_format=picosvg`

The svg file:

```svg
<svg width="200" height="200" viewBox="0 0 200 200" version="1.1" xmlns="http://www.w3.org/2000/svg">
    <circle fill="#F49924" cx="100" cy="100" r="100" />
    <ellipse fill="#7E4418" cx="59.545" cy="71.818" rx="15" ry="21.818" />
    <ellipse fill="#7E4418" cx="140.455" cy="71.818" rx="15" ry="21.818" />
</svg>
```

TTX file:
```xml
 <SVG>
    <svgDoc endGlyphID="2" startGlyphID="2">
      <![CDATA[<svg xmlns="http://www.w3.org/2000/svg" version="1.1"><defs/><g id="glyph2" transform="matrix(6 0 0 6 0 -950)"><path d="M200,100 A100 100 0 1 1 0,100 A100 100 0 1 1 200,100 Z" fill="#F49924"/><path d="M74.545,71.818 A15 21.818 0 1 1 44.545,71.818 A15 21.818 0 1 1 74.545,71.818 Z" fill="#7E4418" id="glyph2::1"/><use href="#glyph2::1" x="80.91"/></g></svg>]]>
    </svgDoc>
  </SVG>
```

I think Safari does not support `<use href="#glyph2::1" x="80.91"/>`, so the ellipse on the right is missing.

![image](https://user-images.githubusercontent.com/2784308/114188821-17d8a300-997c-11eb-8493-be4727508679.png)

## Expect:
![image](https://user-images.githubusercontent.com/2784308/114189132-6dad4b00-997c-11eb-97c2-e3c8bdd52d9e.png)

## Keynote

![image](https://user-images.githubusercontent.com/2784308/114216702-d3f49680-9999-11eb-80e0-4c64168cd2ad.png)

anthrotype:
Interesting, thanks for reporting. If it works on FireFox then this is a Safari bug. You should probably file an issue with Apple.
I wonder if it has anything to do with the way the id string is formatted, e.g. the presence of `::` characters. Could you see if you change the id to something that only contains letters or numbers it works then? This is just a guess of course

yisibl:
I changed :: characters in ttx and rebuilt it to ttf, but it still doesn't work.

Although this is a bug in Safari, can we add an option not to generate use?

yisibl:
```xml
[...]

PR #267, opened 2021-04-09
fix transform for untouchedsvg output

anthrotype:
When we build OT-SVG with `--color_format=untouchedsvg` we are not correctly positioning the SVG in the OT-SVG space. We need to use the same method that we use for the `picosvg` output for mapping from viewBox to OT-SVG space.

Originally reported by @yisibl in https://github.com/googlefonts/nanoemoji/issues/266#issuecomment-816715374

He confirmed (in private chat) that this fixes the positioning issue in Safari/Firefox.


anthrotype:
It may also be that adding a `transform` directly to the root `<svg>` element like we are currently doing was the cause of the issue that @yisibl reported (I believe it was with the Safari browser).

In any case, adding a wrapping `<g>` group with a glyph `id` and a `transform` attributes is also what we are already doing for the `--color_format-picosvg` and that works, so for consistency we should do the same for `untouchedsvg` like in this PR.

anthrotype:
Plus, this PR also makes sure that the `untouchedsvg` output will behave the same way as `picosvg` and `colr_1` outputs with regards to the scaling/positioning of the SVG in the (ascender-descender) height and the glyph advance width. The naive translation that we are currently doing for  `untouchedsvg` did not take that into account.

rsheeter (approved):

yisibl:
Thanks for such a quick fix.

anthrotype:
Thank you for reporting and testing this!

PR #270, opened 2021-04-13
fix: the href attribute of the use element is changed to xlink:href

yisibl:
The `<use>` element must use the `xlink:href` attribute, not `href`. Otherwise it will not take effect in Safari/Mac/iOS.

Although the SVG2 specification recommends using href instead of `xlink:href`, the svg-native specification says:

> deprecated xlink attributes except for 'xlink:href'
> https://svgwg.org/specs/svg-native/#commonattributes

Fixed: https://github.com/googlefonts/nanoemoji/issues/266

Add and update the test case later.

### DIFF
--- a/src/nanoemoji/svg.py
+++ b/src/nanoemoji/svg.py
@@ -20,6 +20,7 @@ from fontTools import ttLib
 from lxml import etree  # pytype: disable=import-error
 from nanoemoji.color_glyph import ColorGlyph, PaintedLayer
 from nanoemoji.disjoint_set import DisjointSet
+from picosvg.svg_meta import xlinkns
 
 from nanoemoji.paint import (
     Extend,
@@ -305,9 +306,6 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
     upem_to_vbox = vbox_to_upem.inverse()
 
     # copy the shapes into our svg
-    XLINK_NAMESPACE = "http://www.w3.org/1999/xlink"
-    XLINK = "{%s}" % XLINK_NAMESPACE
-
     for painted_layer in color_glyph.painted_layers:
         reuse_key = _inter_glyph_reuse_key(view_box, painted_layer)
         if reuse_key not in reuse_cache.shapes:
@@ -320,13 +318,10 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
             for reuse in painted_layer.reuses:
                 _ensure_has_id(el)
 
-                # The xlink:href attribute must be added to the <use> element, 
-                # otherwise it will not take effect in Safari/Mac/iOS.
-                svg_use = etree.SubElement(
-                    svg_g, "use",
-                )
-
-                svg_use.attrib[XLINK + "href"] = f'#{el.attrib["id"]}'
+                svg_use = etree.SubElement(svg_g, "use", nsmap={"xlink": xlinkns()})
+                # Add xlink:href attribute.
+                # https://svgwg.org/specs/svg-native/#commonattributes
+                svg_use.attrib[f"{{{xlinkns()}}}href"] = f'#{el.attrib["id"]}'
                 tx, ty = reuse.gettranslate()
                 if tx:
                     svg_use.attrib["x"] = _ntos(tx)
@@ -339,8 +334,8 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
         else:
             el = reuse_cache.shapes[reuse_key]
             _ensure_has_id(el)
-            svg_use = etree.SubElement(svg_g, "use", nsmap={"xlink": XLINK_NAMESPACE})
-            svg_use.attrib[XLINK + "href"] = f'#{el.attrib["id"]}'
+            svg_use = etree.SubElement(svg_g, "use", nsmap={"xlink": xlinkns()})
+            svg_use.attrib[f"{{{xlinkns()}}}href"] = f'#{el.attrib["id"]}'
 
 
 def _ensure_ttfont_fully_decompiled(ttfont: ttLib.TTFont):
@@ -415,14 +410,14 @@ def _picosvg_docs(
         reuse_cache = ReuseCache()
         # establish base svg, defs
         svg = SVG.fromstring(
-            r'<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><defs/></svg>'
-        , False)
+            r'<svg version="1.1" xmlns="http://www.w3.org/2000/svg"><defs/></svg>'
+        )
 
         for color_glyph in (color_glyphs[g] for g in group):
             _add_glyph(svg, color_glyph, reuse_cache)
 
         gids = tuple(color_glyphs[g].glyph_id for g in group)
-        doc_list.append((svg.tostring(False), min(gids), max(gids)))
+        doc_list.append((svg.tostring(), min(gids), max(gids)))
 
     return doc_list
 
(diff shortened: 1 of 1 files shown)


src/nanoemoji/svg.py
```diff
@@ -20,6 +20,8 @@
 from lxml import etree  # pytype: disable=import-error
 from nanoemoji.color_glyph import ColorGlyph, PaintedLayer
 from nanoemoji.disjoint_set import DisjointSet
+from absl import logging
```
anthrotype: unused

src/nanoemoji/svg.py
```diff
@@ -304,6 +306,9 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
     upem_to_vbox = vbox_to_upem.inverse()
 
     # copy the shapes into our svg
+    XLINK_NAMESPACE = "http://www.w3.org/1999/xlink"
```
anthrotype, line 308: you can `from picosvg.svg_meta import xlinkns`, then use as `xlinkns()`

src/nanoemoji/svg.py
```diff
@@ -304,6 +306,9 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
...
+    XLINK = "{%s}" % XLINK_NAMESPACE
```
anthrotype, line 309: we don't need this. When you want to create an xlink attribute, you do e.g. `f"{{{xlinkns()}}}href` (that's also how picosvg does)

src/nanoemoji/svg.py
```diff
@@ -315,8 +319,14 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
             reuse_cache.shapes[reuse_key] = el
             for reuse in painted_layer.reuses:
                 _ensure_has_id(el)
-                svg_use = etree.SubElement(svg_g, "use")
-                svg_use.attrib["href"] = f'#{el.attrib["id"]}'
+
+                # The xlink:href attribute must be added to the <use> element, 
+                # otherwise it will not take effect in Safari/Mac/iOS.
```
anthrotype, line 324: remove the comment, the use of unqualified `href` without the xlink namespace was our own misunderstanding of the spec. Even picosvg always uses `xlink:href` for `<use>` element. It's the way it is, no need to mention Safari, etc

src/nanoemoji/svg.py
```diff
@@ -405,14 +415,15 @@ def _picosvg_docs(
...
-        )
+            r'<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><defs/></svg>'
+        , False)
 
         for color_glyph in (color_glyphs[g] for g in group):
             _add_glyph(svg, color_glyph, reuse_cache)
 
         gids = tuple(color_glyphs[g].glyph_id for g in group)
-        doc_list.append((svg.tostring(), min(gids), max(gids)))
+        doc_list.append((svg.tostring(False), min(gids), max(gids)))
```
yisibl, line 425: I found that if the fix_xlink_ns option is not added, xmlns:xlink cannot be added.

src/nanoemoji/svg.py
```diff
@@ -315,8 +319,14 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
             reuse_cache.shapes[reuse_key] = el
             for reuse in painted_layer.reuses:
                 _ensure_has_id(el)
-                svg_use = etree.SubElement(svg_g, "use")
-                svg_use.attrib["href"] = f'#{el.attrib["id"]}'
+
+                # The xlink:href attribute must be added to the <use> element, 
+                # otherwise it will not take effect in Safari/Mac/iOS.
+                svg_use = etree.SubElement(
+                    svg_g, "use",
```
anthrotype, line 326: no need to split into multiple lines

src/nanoemoji/svg.py
```diff
@@ -315,8 +319,14 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
...
+                )
+
+                svg_use.attrib[XLINK + "href"] = f'#{el.attrib["id"]}'
```
anthrotype, line 329:
```suggestion
                svg_use.attrib[f"{{{xlinkns()}}}href"] = f'#{el.attrib["id"]}'
```

src/nanoemoji/svg.py
anthrotype, line 425: we don't need to have an option to keep xmlns, that's added or removed automatically depending on whether any xlink attribute is actually used. We can close https://github.com/googlefonts/nanoemoji/pull/270

```suggestion
        doc_list.append((svg.tostring(), min(gids), max(gids)))
```

src/nanoemoji/svg.py
```diff
@@ -405,14 +415,15 @@ def _picosvg_docs(
         reuse_cache = ReuseCache()
         # establish base svg, defs
         svg = SVG.fromstring(
-            r'<svg version="1.1" xmlns="http://www.w3.org/2000/svg"><defs/></svg>'
-        )
+            r'<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><defs/></svg>'
+        , False)
```
anthrotype, line 419:
```suggestion
        )
```

src/nanoemoji/svg.py
```diff
@@ -405,14 +415,15 @@ def _picosvg_docs(
         reuse_cache = ReuseCache()
         # establish base svg, defs
         svg = SVG.fromstring(
-            r'<svg version="1.1" xmlns="http://www.w3.org/2000/svg"><defs/></svg>'
-        )
+            r'<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><defs/></svg>'
```
anthrotype, line 418: don't manually add the xlink ns, it's added automatically depending on the usage of xlink attributes when you do `tostring`. Just add the xlink:href to the use element, and we're done

```suggestion
            r'<svg version="1.1" xmlns="http://www.w3.org/2000/svg"><defs/></svg>'
```

src/nanoemoji/svg.py
```diff
@@ -315,8 +317,11 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
             reuse_cache.shapes[reuse_key] = el
             for reuse in painted_layer.reuses:
                 _ensure_has_id(el)
-                svg_use = etree.SubElement(svg_g, "use")
-                svg_use.attrib["href"] = f'#{el.attrib["id"]}'
+
+                svg_use = etree.SubElement(svg_g, "use", nsmap={"xlink": xlinkns()})
```
yisibl, line 321: Currently, if nsmap is deleted, the wrong namespace will be generated.

src/nanoemoji/svg.py
yisibl: OK, removed.

src/nanoemoji/svg.py
yisibl: Done

src/nanoemoji/svg.py
yisibl: Like this:
```svg
<svg xmlns="http://www.w3.org/2000/svg" version="1.1">
    <defs />
    <g id="glyph2" transform="matrix(6 0 0 6 0 -950)">
        <path d="M200,100 A100 100 0 1 1 0,100 A100 100 0 1 1 200,100 Z" fill="#F49924" />
        <path d="M74.545,71.818 A15 21.818 0 1 1 44.545,71.818 A15 21.818 0 1 1 74.545,71.818 Z" fill="#7E4418" id="glyph2::1" />
        <use xmlns:ns0="http://www.w3.org/1999/xlink" ns0:href="#glyph2::1" x="80.91" />
    </g>
</svg>
```

PR #271, opened 2021-04-13
SVG1.1 requires <use> href attr to be qualified as xlink:href

anthrotype:
We were using the unqualified `href` attribute notation with the `<use>` element, which is [only supported in SVG2](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xlink:href). For `<use>` to work properly on Safari and Adobe apps we need to declare the `xlink` namespace and qualify the attribute as `xlink:href`

Fixes https://github.com/googlefonts/nanoemoji/issues/264 and https://github.com/googlefonts/nanoemoji/issues/266

Supersedes https://github.com/googlefonts/picosvg/pull/207 and https://github.com/googlefonts/nanoemoji/pull/270

Many thanks @yisibl for helping debugging and fixing this!

PR #272, opened 2021-04-13
strip empty <defs/> when no gradients in --color_format=picosvg output

anthrotype:
@yisibl =>  https://github.com/googlefonts/nanoemoji/pull/271#discussion_r612347962


yisibl:
Great, when there are more glyphs like this, the font file size can be reduced significantly.

Issue #268, opened 2021-04-09
Svg font does not support gradientTransform in Safari


yisibl:
`nanoemoji --color_format=untouchedsvg`

```svg
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 128 128">
  <defs>
    <linearGradient id="b" gradientUnits="userSpaceOnUse" x1="98.95" x2="123.82" y1="68.26" y2="93.14">
      <stop offset="0" stop-color="#2d50ff"/>
      <stop offset="1" stop-color="#61bef1"/>
    </linearGradient>
    <linearGradient id="e" gradientTransform="scale(-1 1)" gradientUnits="userSpaceOnUse" x1="-29.06" x2="-4.18" y1="68.26" y2="93.14">
      <stop offset="0" stop-color="#2d50ff"/>
      <stop offset="1" stop-color="#61bef1"/>
    </linearGradient>
  </defs>
  <path fill="url(#b)" d="M108.49,93.14 C104.25,88.9 97.25,70 99,68.26 S119.64,73.56 123.87,77.8 A10.84 10.84 0 1 1 108.54,93.14 Z"/>
  <path fill="url(#e)" d="M19.51,93.14 C23.75,88.9 30.75,70 29.05,68.26 S8.41,73.56 4.18,77.8 A10.84 10.84 0 0 0 19.51,93.14 Z"/>
</svg>
```

## Safari
![image](https://user-images.githubusercontent.com/2784308/114214859-73fcf080-9997-11eb-8b0a-157a6cd36ee9.png)

## Firefox
![image](https://user-images.githubusercontent.com/2784308/114216121-0fdb2c00-9999-11eb-90a4-405c66694bec.png)

## Adobe Photoshop

![image](https://user-images.githubusercontent.com/2784308/114214867-77907780-9997-11eb-9d84-702130236f9e.png)

yisibl:
Thanks @litherum for providing a solution in https://twitter.com/Litherum/status/1381638423237156868

anthrotype:
actually, it's more a workaround than a "solution".

[...]

PR #273, opened 2021-04-13
work around Safari's bug when gradientTransform contains "involutory" matrix

anthrotype:
Fixes https://github.com/googlefonts/nanoemoji/issues/268

Safari currently rejects a gradient if `gradientTransform` attribute contains a matrix whose inverse is equal to itself (aka [involutory matrix](https://en.wikipedia.org/wiki/Involutory_matrix)). These are perfectly valid, non-singular invertible transformations which happen to have an inverse which is equal to themselves (notably, reflection across x/y axis). Safari assumes they are non-invertible and incorrectly rejects the whole gradient. The [recommendation from the WebKit maintainer](https://github.com/litherum) is to hack the matrix by a small, invisible amount enough to trick Safari.
I heard this bug has already been fixed internally but will take time until it reaches all users. Until then, we have no choice but doing this.

### DIFF
--- a/src/nanoemoji/svg.py
+++ b/src/nanoemoji/svg.py
@@ -40,6 +40,10 @@ from picosvg.svg_types import SVGPath
 from typing import MutableMapping, NamedTuple, Optional, Sequence, Tuple, Union
 
 
+# topicosvg()'s default
+_DEFAULT_ROUND_NDIGITS = 3
+
+
 class InterGlyphReuseKey(NamedTuple):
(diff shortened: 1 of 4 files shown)


anthrotype:
Note this will only affect `--color_format=picosvg` (in which we rewrite the svg shapes), but not the `untouchedsvg` format. As we don't touch the latter, it's the responsibility of the user to deal with this quirk in Safari if they wish to.

PR #280, opened 2021-04-19
use 'ccmp' feature for required ligature substitution

### LINKED ISSUE #279: Ligature changed from rlig feature to ccmp
Different platforms or browsers have some differences in ligature support([Test](http://unifraktur.sourceforge.net/testcases/enable_opentype_features/)). Let us discuss what issues should be paid attention to in emoji ligatures.

noto-emoji and noto-emoji-svg both use ccmp by default

https://github.com/googlefonts/noto-emoji/blob/41ae6686ace1453b432ac907a165428f2e1ad54e/NotoColorEmoji.tmpl.ttx.tmpl#L311

https://github.com/adobe-fonts/noto-emoji-svg/blob/ed5c78c8e3d46fdf8dbf7532875aaabfdcdc5c3a/GSUB.fea#L4

anthrotype:
instead of 'rlig', for wider compatibility. NotoColorEmoji.ttf does the same
Fixes #279

### DIFF
--- a/src/nanoemoji/features.py
+++ b/src/nanoemoji/features.py
@@ -20,14 +20,17 @@
 from nanoemoji.glyph import glyph_name
 
 
-def generate_fea(rgi_sequences):
-    # Generate rlig feature with ligature lookup for multi-codepoint RGIs
+DEFAULT_GSUB_FEATURE_TAG = "ccmp"
+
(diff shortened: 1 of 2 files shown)


yisibl:
looks great!

PR #289, opened 2021-05-14
Don't reuse components if they overlap and have a reversing transform

### LINKED ISSUE #287: winding issue when intra-glyph reused shapes overlap
In `nanoemoji.color_glyph._painted_layers`, we group consecutive paths with the same paint and the same "normalized" version (i.e. one that compares == even if offset, scaled, rotated, etc.) in the same `PaintedLayer`, which later on gets encoded as a single `glyf` composite glyph, the reused paths encoded as components of the first one with a given transform.

But when these components overlap with each other and the transform is such that the winding contour direction gets flipped (negative determinant), then we can see glitches such as this:

Screenshot 2021-05-10 at 17 30 17

This is more visible when we zoom in on one of the stars (which in this particular flag are comprised of many triangles overlapping one another):

Screenshot 2021-05-10 at 17 30 29

I think we should avoid the intra-glyph reuse if the components' paths overlap with one another and the transform flips the winding direction.

We can use `picosvg.svg_types.intersection` function to check if any pair of paths that could be reused do overlap (their intersection is not empty), and if they do, we revert to encoding them as distinct PaintedLayers (which will get encoded as distinct contour glyfs)

anthrotype:
Fixes #287 

Before (composite glyph with overlapping components and reversing transforms):

Screenshot 2021-05-14 at 19 45 28

```xml
    <TTGlyph name="glyph00005" xMin="495" yMin="163" xMax="546" yMax="621">
      <component glyphName="glyph00006" x="0" y="0" flags="0x4"/>
      <component glyphName="glyph00006" x="535" y="1419" scalex="0.7314" scale01="-0.7173" scale10="-0.6503" scaley="-0.7214" flags="0x4"/>
      <component glyphName="glyph00006" x="535" y="1003" scalex="0.7314" scale01="-0.7173" scale10="-0.6503" scaley="-0.7214" flags="0x4"/>
      <component glyphName="glyph00006" x="0" y="-416" flags="0x4"/>
    </TTGlyph>

    <TTGlyph name="glyph00006" xMin="495" yMin="579" xMax="530" yMax="618">
      <contour>
        <pt x="530" y="579" on="1"/>
        <pt x="495" y="615" on="1"/>
        <pt x="517" y="618" on="1"/>
      </contour>
      <instructions/>
    </TTGlyph>
```

After (simple contour glyphs):

Screenshot 2021-05-14 at 19 45 36

```xml
    <TTGlyph name="glyph00005" xMin="495" yMin="579" xMax="530" yMax="618">
      <contour>
        <pt x="530" y="579" on="1"/>
        <pt x="495" y="615" on="1"/>
        <pt x="517" y="618" on="1"/>
      </contour>
      <instructions/>
    </TTGlyph>

    <TTGlyph name="glyph00006" xMin="497" yMin="602" xMax="546" yMax="621">
      <contour>
        <pt x="546" y="621" on="1"/>
        <pt x="497" y="620" on="1"/>
        <pt x="510" y="602" on="1"/>
      </contour>
      <instructions/>
    </TTGlyph>

    <TTGlyph name="glyph00007" xMin="497" yMin="187" xMax="546" yMax="205">
      <contour>
        <pt x="546" y="205" on="1"/>
        <pt x="497" y="204" on="1"/>
        <pt x="510" y="187" on="1"/>
      </contour>
      <instructions/>
    </TTGlyph>

    <TTGlyph name="glyph00008" xMin="495" yMin="163" xMax="530" yMax="203">
      <contour>
        <pt x="530" y="163" on="1"/>
        <pt x="495" y="199" on="1"/>
        <pt x="517" y="203" on="1"/>
      </contour>
      <instructions/>
    </TTGlyph>
```

### DIFF
--- a/src/nanoemoji/color_glyph.py
+++ b/src/nanoemoji/color_glyph.py
@@ -14,7 +14,7 @@
 
 import math
 from absl import logging
-from itertools import chain, groupby
+from itertools import chain, groupby, combinations
 from lxml import etree  # type: ignore
 from nanoemoji.colors import Color
 from nanoemoji.config import FontConfig
@@ -33,9 +33,15 @@ from picosvg.svg_meta import number_or_percentage
 from picosvg.svg_reuse import normalize, affine_between
 from picosvg.svg_transform import Affine2D
 from picosvg.svg import SVG
-from picosvg.svg_types import SVGPath, SVGLinearGradient, SVGRadialGradient
+from picosvg.svg_types import (
+    SVGPath,
+    SVGLinearGradient,
+    SVGRadialGradient,
+    intersection,
+)
 from typing import Generator, NamedTuple, Optional, Sequence, Tuple
 import ufoLib2
+import pathops
 
 
 def _scale_viewbox_to_font_metrics(
@@ -286,6 +292,15 @@ def _in_glyph_reuse_key(
     )
 
 
+def _intersect(path1: SVGPath, path2: SVGPath) -> bool:
+    # Try computing intersection using pathops; if for whatever reason it fails
+    # (probably some bug) then be on the safe side and assume we do have one...
+    try:
+        return bool(intersection((path1, path2)))
+    except pathops.PathOpsError:
+        return True
+
+
 def _painted_layers(
     debug_hint: str,
     config: FontConfig,
@@ -307,14 +322,14 @@ def _painted_layers(
         key=lambda s: _in_glyph_reuse_key(debug_hint, config, picosvg, s, glyph_width),
     ):
         paths = list(paths)
-        transforms = ()
+        transforms = [Affine2D.identity()]
         if len(paths) > 1:
-            transforms = tuple(
+            transforms.extend(
                 affine_between(paths[0], p, config.reuse_tolerance) for p in paths[1:]
             )
 
         success = True
-        for path, transform in zip(paths[1:], transforms):
+        for path, transform in zip(paths[1:], transforms[1:]):
             if transform is None:
                 success = False
                 error_msg = (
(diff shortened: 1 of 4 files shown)


src/nanoemoji/color_glyph.py
```diff
@@ -286,6 +292,15 @@ def _in_glyph_reuse_key(
     )
 
 
+def _intersect(path1: SVGPath, path2: SVGPath) -> bool:
+    # Try computing intersection using pathops; if for whatever reason it fails
+    # (probably some bug) then be on the safe side and assume we do have one...
+    try:
+        return bool(intersection((path1, path2)))
+    except pathops.PathOpsError:
+        return True
```
rsheeter, line 301: I think we should at least log this; shouldn't really be happening

src/nanoemoji/color_glyph.py
```diff
@@ -326,8 +341,22 @@ def _painted_layers(
...
+            # creates holes where they overlap as nonzero fill rule is applied by
+            # TrueType renderer: https://github.com/googlefonts/nanoemoji/issues/287
+            reverses_direction = [t.determinant() < 0 for t in transforms]
+            success = not any(
+                (
+                    (reverses_direction[i] or reverses_direction[j])
+                    and _intersect(paths[i], paths[j])
+                )
+                for i, j in combinations(range(len(paths)), 2)
+            )
```
rsheeter, line 356: I think we should log when this kicks in

rsheeter (approved):
Would love to know how often this kicks in for a Noto or Twemoji build

src/nanoemoji/color_glyph.py
anthrotype: it's really unlikely that we ever encounter this, but I didn't want to risk it given recent breakages with pathops.simplify. I added the logging message

src/nanoemoji/color_glyph.py
anthrotype: done

anthrotype:
> Would love to know how often this kicks in for a Noto or Twemoji build

sure, will merge this and I'll let you know after I try.

anthrotype:
I'll fix the test failures in a subsequent PR (due to changes in latest fonttools colorLib, no longer adding an empty LayersV1List which is optional in COLRv1).

anthrotype:
> Would love to know how often this kicks in for a Noto or Twemoji build

I counted 287 emojis that are affected by this change, out of a total 3573 in noto-glyf_colr_1.ttf

I used this script to parse the nanoemoji build log, grepping for the relevant logging message

```python
import fileinput
import re

filename_RE = re.compile(
    r".*(emoji_u\w+\.svg).*overlap and have a reversing transform.*"
)

filenames = set()
for ln in fileinput.input():
    m = filename_RE.match(ln)
    if m:
        filenames.add(m.group(1))

print(len(filenames))
print(*sorted(filenames), sep="\n")
```

src/nanoemoji/color_glyph.py
anthrotype: > shouldn't really be happening

I just checked and this PathOpsError never happens when building Noto Color Emoji. If we like, we can let the error propagate instead

PR #296, opened 2021-06-29
Add support for layer graphs

rsheeter:
Prepare for the output that https://github.com/googlefonts/picosvg/pull/223 will cause.

Note that I froze the picosvg version on main, released main, and am unfreezing here. Imperfect but good enough I think.

Edit: more context per request.

Thankfully a lot is test files. That said, while I don't know you need to recompute every value in them looking at the broad patterns of change in the test files does seem valuable. Here's my first swing at a readers guide, may or may not be helpful. I also tentatively put time on calendar tomorrow but no worries if it doesn't work.

**various _colr_1.ttx files**
note broad strokes: fewer glyphs, fewer components, more COLR transform based reuse

**config.py** (minor change)
it's a nightmare to review diffs of svg ttx because the svg is not pretty. Make an option to make it pretty.

tests/clocks_picosvg.ttx
pretty printed svgDoc, why didn't I do that sooner?!

**paint.py**
now we have a graph it's helpful to be able to wander around it
so we have traversal APIs
notably, providing the current transform and the sequence of nodes to get to here

**tests/color_glyph_test.py**
informative, I think/hope, because it shows the graphs we're building and that may put rest of code into context
test w/group_opacity.svg exhibits composite being emitted

**svg.py**
it's conceptually unchanged
_glyph_groups makes sure if sharing of shapes is likely we use the same svg document
we've upgraded reuse to use the normalize/affine_between mechanism; it should work somewhat better
add_glyph is notably incapable of handling anything other than paint glyph

**color_glyph.py**
_painted_layers is the main change
loop from leaf to root because that Paint builds more naturally that way (leaf is referred to by immutable parent)
shapes more or less handled the same
groups now push PaintComposite
anything else is unsupported. Silently ... maybe that should throw.

_mutating_traverse
helper for wandering around a color glyph altering Paint's

**write_font.py**
heavy user of the traverse apis we put into paint.py

GlyphReuseCache is a good starting point; heart of updated reuse computation

_migrate_paths_to_ufo_glyphs uses it to transfer content into the ufo structures
we replace limited componentization with global Transform of Glyph based reuse
...in case of consecutive shapes in same paint we might want to put back components
see tests/rect_colr_1.ttx for hwo this plays out; fewer TTGlyph and more COLR transformation [waiting to exploit the fancy new small transforms]

_glyf_ufo
builds a rather dumbed down 'glyf' based output

_colr0_layers
I should switch the .format check to the isinstance one
similar to _glyf_ufo, dumbed down COLR v0

_bounds
tries to correctly accumulate bounds factoring in that some of our glyph are reused with an additional transform

**tests/clocks_glyf.ttx**
interesting because it exhibits use of component w/transform


src/nanoemoji/color_glyph.py
```diff
@@ -472,10 +510,23 @@ def transform_for_otsvg_space(self):
...
     def colors(self):
         """Set of Color used by this glyph."""
-        return set(chain.from_iterable([p.colors() for p in self.paints()]))
+        all_colors = set()
+        self.traverse(lambda paint: all_colors.update(paint.colors()))
+        return all_colors
+
+    def traverse(self, visitor):
+        def _traverse_callback(paint):
+            updated_paint = visitor(paint)
```
rsheeter, line 521: either don't capture updated_paint or assign to _ per @anthrotype

src/nanoemoji/svg.py
```diff
@@ -322,45 +331,73 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
...
+    for root in color_glyph.painted_layers:
+        for context in root.breadth_first():
+            if any(c == context.path[: len(c)] for c in complete_paths):
+                continue
+
+            parent_el = svg_g
+            if context.path:
+                parent_el = context.path[-1]
+
+            if isinstance(context.paint, PaintGlyph):
```
rsheeter, line 345: should handle PaintComposite here too

src/nanoemoji/svg.py
```diff
@@ -322,45 +331,73 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
...
+                            svg_use.attrib["y"] = _ntos(ty)
+                        transform = transform.translate(-tx, -ty)
+                        if transform != Affine2D.identity():
+                            svg_use.attrib["transform"] = _svg_matrix(transform)
+
+                        created_use = True
+
+                if not created_use:
+                    el = to_element(SVGPath(d=paint_glyph.glyph))
+                    _apply_paint(
```
rsheeter, line 392: ...should be checking transform here? - sanity check we have a test w/transforms

src/nanoemoji/svg.py
rsheeter: subset of functionality that we actually emit

src/nanoemoji/color_glyph.py
```diff
@@ -332,55 +348,50 @@ def _painted_layers(
...
-                debug_hint, paths, transforms
+        if context.is_group():
+            # flush the current shapes into a new group
+            opacity = float(context.element.get("opacity"))
+            assert 0.0 < opacity < 1.0, f"{context.path} should be transparent"
+            assert len(nodes) > 1, f"{context.path} should have 2+ children"
+            assert {"opacity"} == set(
+                context.element.attrib.keys()
+            ), f"{context.path} only attribute should be opacity. Found {context.element.attrib.keys()}"
+            paint = PaintComposite(
```
rsheeter, line 382: emit a font and try it in Chrome

src/nanoemoji/write_font.py
```diff
@@ -229,51 +236,114 @@ def _next_name(ufo: ufoLib2.Font, name_fn) -> str:
     return name_fn(i)
 
 
-def _create_glyph(color_glyph: ColorGlyph, painted_layer: PaintedLayer) -> Glyph:
+class ReuseResult(NamedTuple):
+    glyph_name: str
+    transform: Affine2D
+
+
+class GlyphReuseCache:
```
rsheeter, line 244: @anthrotype notes that there may be a threshold where cost of reuse is higher than cost of emitting the glyph (think triangle). Maybe don't fix here but file an issue to investigate?

rsheeter: Tracking in #302

src/nanoemoji/write_font.py
```diff
@@ -229,51 +236,114 @@ def _next_name(ufo: ufoLib2.Font, name_fn) -> str:
...
+        self._config = config
+        self._known_glyphs = set()
+        self._reusable_paths = {}
+
+    def try_reuse(self, path: str) -> Optional[ReuseResult]:
+        """Try to reproduce path as the transformation of another glyph.
+
+        Path is expected to be in font units.
+
+        Returns (glyph name, transform) if possible, (input path, identity) if not.
```
anthrotype, line 255:
```suggestion
        Returns (glyph name, transform) if possible, `None` if not.
```

src/nanoemoji/write_font.py
```diff
@@ -229,51 +236,114 @@ def _next_name(ufo: ufoLib2.Font, name_fn) -> str:
...
+        if norm_path not in self._reusable_paths:
+            print(f"  not in reusable {norm_path}")
+            return None
+
+        glyph_name, glyph_path = self._reusable_paths[norm_path]
+        affine = affine_between(
+            SVGPath(d=glyph_path), SVGPath(d=path), self._config.reuse_tolerance
+        )
+        if affine is None:
+            print(f"  affine_between failed {norm_path}")
```
anthrotype, line 277: it used to be a logging.warning when we grouped a path but no `affine_between` could be found (signalling a potential picosvg bug), maybe we keep that?

BTW, all print() must go ;)

src/nanoemoji/write_font.py
```diff
@@ -296,35 +366,44 @@ def _draw_glyph_extents(
...
-def _glyf_ufo(ufo, color_glyphs):
+def _glyf_ufo(
+    config: FontConfig, ufo: ufoLib2.Font, color_glyphs: MutableSequence[ColorGlyph]
+):
     # glyphs by reuse_key
-    glyphs = {}
-    reused = set()
-    for color_glyph in color_glyphs:
+    glyph_cache = GlyphReuseCache(config)
+    glyph_uses = defaultdict(int)
```
anthrotype, line 374: make it a `collections.Counter`?

src/nanoemoji/write_font.py
```diff
@@ -296,35 +366,44 @@ def _draw_glyph_extents(
...
-            parent_glyph.components.append(Component(baseGlyph=glyph.name))
 
+        # generate glyphs for PaintGlyph's and assign glyph names
+        color_glyphs[i] = color_glyph = _migrate_paths_to_ufo_glyphs(
+            color_glyph, glyph_cache
+        )
+
+        for root in color_glyph.painted_layers:
+            for context in root.breadth_first():
+                # For 'glyf' just dump anything that isn't a PaintGlyph
```
anthrotype, line 391: should we log a warning if that isn't the case?

tests/write_font_test.py
```diff
@@ -222,15 +226,15 @@ def test_vertical_metrics(ascender, descender, linegap):
...
         # we work around it by nudging one matrix component by an invisible amount
         # https://github.com/googlefonts/nanoemoji/issues/268
         (
             ("involutory_matrix.svg",),
             "involutory_matrix_picosvg.ttx",
-            {"color_format": "picosvg"},
+            {"color_format": "picosvg", "pretty_print": True},
         ),
         # Check that we do _not_ make composite glyphs with reused paths if
         # the latter overlap and the transform for the shape reuse is such
```
anthrotype, line 240: since we are no longer creating glyf composites (with potentially overlapping components that may create winding issues thus is safer to decompose) but reusing PaintGlyphs via PaintTransforms, the comment above about transformed_components_overlap.ttx may be outdated and the whole test probably no longer relevant?
Actually i'd be curious to know whether reusing "components" as PaintTransform-ed PaintGlyphs (as opposed to traditional glyf composites) incur in the same overlap issues (I suspect not, as each coloured layer is composited over the previous one)..

src/nanoemoji/write_font.py
rsheeter: I can ... concern was there isn't really anything to be done about it. `glyf` just can't produce all the things COLR can.

anthrotype (approved):
LGTM, pending rebase

PR #298, opened 2021-06-29
update COLRv1 struct names to match latest fonttools and draft spec

anthrotype:
Matches fonttools/fonttools#2346 (yet to be released, so tests will fail until then)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ setup(
     include_package_data=True,
     install_requires=[
         "absl-py>=0.9.0",
-        "fonttools[ufo]>=4.23.1",
+        "fonttools[ufo]>=4.25.0",
         "importlib_resources>=3.3.0; python_version < '3.9'",
         "lxml>=4.0",
         "ninja>=1.10.0.post1",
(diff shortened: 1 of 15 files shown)


rsheeter (approved):
LGTM, ideally lets update the required version of fonttools too (assuming release coming shortly)

anthrotype:
yeah, I will cut a fonttools release and bump the required version. But first, it's better to do the remaining otData.py updates (e.g. https://github.com/googlefonts/colr-gradients-spec/pull/307)

anthrotype:
and there's also the new [PaintTransform with indirect Affine2x3 at an offset](https://github.com/googlefonts/colr-gradients-spec/pull/305) which I still need to wrap in otData, makes sense to do all that before tagging fonttools

rsheeter:
+1 to making all the changes in FT in one fell swoop. COLRv1 should settle down after this week.

anthrotype:
> +1 to making all the changes in FT in one fell swoop

Actually, given that freetype2 main branch currently only has PaintScale and the renumbered formats (basically this https://github.com/fonttools/fonttools/pull/2348) and that Chrome Canary automatically rolls the latter, and that we won't get the rest of the spec changes implemented in FT2 before Dominik is back from annual leave in a couple weeks, it makes sense that I cut a fonttools release as is, before we merge in the rest of the changes (indirect Affine2x3, reduced precision specialized transforms, all angles as F2Dot14, centerX/Y as FWORD), so that we keep in sync with FT2 and Canary.

PR #303, opened 2021-07-12
Use new Paint formats

rsheeter:
First steps toward using the new paint formats. Only int16 translate and scale are converted. Builds on #310's update to FT 4.5.2; should merge that first. Until merged may wish to view diff via https://github.com/googlefonts/nanoemoji/compare/ft_452...reduce_prec.

Some rounding seems to occur due to changes in FontTools, for example `'scaleY': 0.333333333` ends up rounded to 0.3333 instead of 0.3333**3**. Presumably this is due to the move from Fixed to F2Dot14.

### DIFF
--- a/tests/colr_to_svg.py
+++ b/tests/colr_to_svg.py
@@ -25,7 +25,7 @@ from nanoemoji.paint import (
     PaintColrGlyph,
     PaintComposite,
     PaintColrLayers,
-    PaintTransform,
+    is_transform,
 )
 from nanoemoji.svg import (
     _svg_matrix,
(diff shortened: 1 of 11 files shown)


tests/colr_to_svg.py
```diff
@@ -220,23 +221,11 @@ def _colr_v1_paint_to_svg(
...
-                        ot_paint.Transform.yy,
-                        ot_paint.Transform.dx,
-                        ot_paint.Transform.dy,
-                    )
-                ),
-                transform,
-            )
+    elif is_transform(ot_paint.Format):
+        paint = Paint.from_ot(ot_paint)
+        transform = Affine2D.product(
```
anthrotype, line 226: we deprecated the `product()` method as it got the operands' order wrong, either use `@` matmul operator or `compose_ltr`

rsheeter: Fixed, ty

tests/svg_colr_svg_test.py
```diff
@@ -46,6 +46,7 @@ def test_svg_to_colr_to_svg(svg_in, expected_svg_out, config_overrides):
         (svg_in,),
     )
     _, ttfont = write_font._generate_color_font(config, glyph_inputs)
+    ttfont.saveXML("-")  # TEMPORARY
```
anthrotype, line 49: remove if TEMPORARY, or just keep if useful for debugging (pytest will capture stdout by default unless assertion fails)

anthrotype (approved):
LGTM

tests/svg_colr_svg_test.py
rsheeter: You were getting good at spotting my prints so I had to start finding new ways to sneak useless console noise past

PR #307, opened 2021-07-19
draw_svg_path: auto-close open sub-paths when drawing svg paths with fonttools pen

anthrotype:
The `draw_svg_path` function which we use to convert SVG paths to FontTools pen commands to create UFO glyphs from them, uses the presence/absence of 'z' to decide whether to call a pen's endPath (for open contours) and closePath (for closed ones). If a filled subpath does not explicily ends with 'z' (which is the case for some of our source files, e.g. 'one-o-clock.svg'), then upon converting it to UFO glyph using the FontTools pen protocol, the subpath is kept 'open' (i.e. the UFO GLIF contour's first point type is set to 'move').
Then, when the UFO glyph is converted to TrueType by ufo2ft with the TTGlyphPointPen, if the starting 'move' point and the last point of such  nominally 'open' contours are on top of each other (i.e. contour actually does a full loop despite being marked as 'open'), the result is a duplicate point at the end/start of the TrueType contour.
We fix this by adding an optional `close_subpath=False` parameter to draw_svg_path function, which when True signals that we want to interpret the all SVG subpaths as if they were closed by an explicit 'z' command, which avoids the duplicate end/start points in the converted TrueType glyphs.
TrueType contours are always filled and automatically closed by an implicit line back to their staring point.

### DIFF
--- a/tests/svg_path_test.py
+++ b/tests/svg_path_test.py
@@ -78,3 +78,10 @@ def test_roundtrip_path_with_pen(d):
     pen = SVGPathPen()
     draw_svg_path(path, pen)
     assert pen.path.d == d
+
+
+def test_draw_svg_close_subpaths():
+    path = SVGPath(d="M0,0 L0,10 L10,10 L10,0 M12,0 L12,10 L22,10 L22,0")
+    pen = SVGPathPen()
+    draw_svg_path(path, pen, close_subpaths=True)
+    assert pen.path.d == "M0,0 L0,10 L10,10 L10,0 Z M12,0 L12,10 L22,10 L22,0 Z"
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ setup(
         "pillow>=7.2.0",
         "regex>=2020.4.4",
         "toml>=0.10.1",
-        "ufo2ft[cffsubr]==2.21.0",  # TEMPORARY VERSION LOCK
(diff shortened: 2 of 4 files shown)


tests/svg_path_test.py
```diff
@@ -78,3 +78,10 @@ def test_roundtrip_path_with_pen(d):
     pen = SVGPathPen()
     draw_svg_path(path, pen)
     assert pen.path.d == d
+
+
+def test_draw_svg_close_subpaths():
+    path = SVGPath(d="M0,0 L0,10 L10,10 L10,0 M12,0 L12,10 L22,10 L22,0")
+    pen = SVGPathPen()
+    draw_svg_path(path, pen, close_subpaths=True)
```
rsheeter, line 86: Do we ever use close subpaths false? - if not perhaps we don't need a parameter

rsheeter (approved):

tests/svg_path_test.py
anthrotype: well, only in the tests. I think I've used the `draw_svg_path` function while fiddling with SVG paths interactively in a font editor, it's still nice to be have to draw the _actual_ svg path with the fonttools pen, even if it's open.

anthrotype: another reason to keep is being able to roundtrip draw_svg_path -> SVGPathPen  for self-testing purposes

PR #310, opened 2021-07-31
Update to FontTools 4.25.2 (latest spec changes)

rsheeter:


src/nanoemoji/paint.py
```diff
@@ -115,17 +118,25 @@ def breadth_first(self) -> Generator[PaintTraverseContext, None, None]:
     def children(self) -> Iterable["Paint"]:
         return ()
 
+    def gettransform(self) -> Optional[Affine2D]:
```
anthrotype, line 121: why did you type the return value as `Optional[Affine2D]` instead of just Affine2D?

src/nanoemoji/paint.py
```diff
@@ -100,13 +103,13 @@ def breadth_first(self) -> Generator[PaintTraverseContext, None, None]:
...
-                transform = Affine2D.compose_ltr(
-                    (
-                        transform,
-                        Affine2D(*context.paint.transform),
-                    )
+            paint_transform = context.paint.gettransform()
+            transform = Affine2D.compose_ltr(
+                (
+                    transform,
+                    Affine2D(*paint_transform),
```
anthrotype, line 110: if the return value from gettransform method is already an Affine2D, why do you need to unpack and construct a new Affine2D?

src/nanoemoji/paint.py
```diff
@@ -115,17 +118,25 @@ def breadth_first(self) -> Generator[PaintTraverseContext, None, None]:
...
     def colors(self):
-        pass
+        for p in self.layers:
+            yield from p.colors()
 
     def to_ufo_paint(self, colors):
-        return [p.to_ufo_paint(colors) for p in self.layers]
+        return {
+            "Format": self.format,
+            "Layers": [p.to_ufo_paint(colors) for p in self.layers],
```
anthrotype, line 138: why are you returning a dict instead of a list? I think colorLib.builder already special-case list of paints as meaning PaintColrLayers,  no? The plist encoded in the UFO.lib will be a bit more compact if you use a list (encoded as plist `<array>` as opposed to a `<dict>` with all the key/values and the implied Format.

src/nanoemoji/paint.py
```diff
@@ -299,6 +312,233 @@ def to_ufo_paint(self, colors):
...
+            "xSkewAngle": self.xSkewAngle,
+            "ySkewAngle": self.ySkewAngle,
+        }
+        return paint
+
+    def children(self) -> Iterable[Paint]:
+        return (self.paint,)
+
+    def gettransform(self) -> Affine2D:
+        raise NotImplementedError()
```
anthrotype, line 511: why NotImplementedError? Affine2D does have `skewx` and `skewy` methods that work on degree angles

src/nanoemoji/paint.py
```diff
@@ -299,6 +312,233 @@ def to_ufo_paint(self, colors):
...
+            "centerX": self.center[0],
+            "centerY": self.center[1],
+        }
+        return paint
+
+    def children(self) -> Iterable[Paint]:
+        return (self.paint,)
+
+    def gettransform(self) -> Affine2D:
+        raise NotImplementedError()
```
anthrotype, line 540: this should be similar to PaintScaleAroundCenter, a combination of translate, skew, (counter-)translate -- or am I missing something?

anthrotype (approved):

src/nanoemoji/paint.py
rsheeter: Sheer incompetance :D

src/nanoemoji/paint.py
rsheeter: I liked the consistency of always using the fieldname: value dict.

src/nanoemoji/paint.py
rsheeter: Quite right, I'm not sure why I didn't fill that in.

src/nanoemoji/paint.py
rsheeter: Added implementation

src/nanoemoji/paint.py
rsheeter: That said, it's just pointlessly making this change larger so I think I'll revert to the old form

PR #311, opened 2021-08-03
Set ufo2ft's colrClipBoxes lib key to build COLRv1 ClipList

anthrotype:
Requires https://github.com/googlefonts/ufo2ft/pull/521 (~~unreleased ufo2ft~~)

### DIFF
--- a/src/nanoemoji/config.py
+++ b/src/nanoemoji/config.py
@@ -87,6 +87,12 @@ flags.DEFINE_bool(
     "Whether to fail or continue with a warning when picosvg cannot compute "
     "affine between paths that normalize the same.",
 )
+flags.DEFINE_integer(
+    "clipbox_quantization",
+    None,
+    "Whether to round COLR clip boxes to multiples of given integer.",
+    lower_bound=1,
+)
 flags.DEFINE_bool(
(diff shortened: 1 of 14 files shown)


src/nanoemoji/config.py
```diff
@@ -87,6 +87,12 @@
     "Whether to fail or continue with a warning when picosvg cannot compute "
     "affine between paths that normalize the same.",
 )
+flags.DEFINE_integer(
+    "clipbox_quantization",
+    None,
+    "Whether to round COLR clip boxes to multiples of given integer.",
```
rsheeter, line 93: Round toward negative/positive infinity hopefully, not the nearest multiple? - should describe so as well

src/nanoemoji/config.py
```diff
@@ -87,6 +87,12 @@
     "Whether to fail or continue with a warning when picosvg cannot compute "
     "affine between paths that normalize the same.",
 )
+flags.DEFINE_integer(
+    "clipbox_quantization",
+    None,
```
rsheeter, line 92: per f2f we should explore >1 default here

rsheeter (approved):

PR #314, opened 2021-08-05
Avoid kerploding due to Affine2x3's that are outside Fixed range.

rsheeter:
Short-term fix to prevent crash. Hoping to later improve the way we go about building glyphs to reduce frequency.

### DIFF
--- a/src/nanoemoji/write_font.py
+++ b/src/nanoemoji/write_font.py
@@ -72,6 +72,10 @@ import ufo2ft
 FLAGS = flags.FLAGS
 
 
+MIN_FIXED = -32768
+MAX_FIXED = 0x7FFFFFFF / (1 << 16)
+
+
 class InputGlyph(NamedTuple):
     filename: Path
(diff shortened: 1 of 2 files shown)


src/nanoemoji/write_font.py
```diff
@@ -260,6 +264,7 @@ def try_reuse(self, path: str) -> Optional[ReuseResult]:
         assert path.startswith("M"), f"{path} doesn't look like a path"
 
         if self._config.reuse_tolerance == -1:
+            print("reuse disabled")
```
anthrotype, line 267: stray print

anthrotype (approved):

src/nanoemoji/write_font.py
anthrotype: I deleted the stray print while rebasing off main branch

anthrotype:
@googlebot I consent.

rsheeter:
I think I need a test that just scans the code for prints!

Issue #316, opened 2021-08-06
Investigate reason for file size increase in COLRv1 since v0.8.3


anthrotype:
cont'd from https://github.com/googlefonts/nanoemoji/issues/312#issuecomment-894177754

> we are seeing an increase in the overall length of the COLR table as a whole -- from 1,054,030 in the original [noto-glyf_colr_1.ttf] font to 1,364,730 in the new one (+310,700 bytes)

In particular, I see that the COLR.LayerList table size increased from 1,010,540 to 1,294,201 (+283,661) bytes. The total number of layers in the LayerList increased from 62782 to 73814 (+11,032).

anthrotype:
Interestingly, the `font["maxp"].numGlyphs` (which counts all the glyphs in the font, including those used as PaintGlyph layers) _decreased_ from 45,459 to 41,793 (-3,666).

On the other hand, the total number of base color glyphs (`COLR.BaseGlyphList.BaseGlyphCount`) is the same in both old and new font: 3,573 (in fact the source files are the same for both).

anthrotype:
I counted the number of unique Paint tables in the old and new font, both total and per-format.

For the old font we have a total of 81,151 paints, subdivided as:

```
{<PaintFormat.PaintGlyph: 10>: 55198,
 <PaintFormat.PaintLinearGradient: 4>: 4939,
 <PaintFormat.PaintColrLayers: 1>: 5828,
 <PaintFormat.PaintSolid: 2>: 3604,
 <PaintFormat.PaintTransform: 12>: 5307,
 <PaintFormat.PaintRadialGradient: 6>: 6275}
```

For the new font, we have a total of 96,063 (**+14,912**) subdivided as:
```
{<PaintFormat.PaintGlyph: 10>: 52497,
 <PaintFormat.PaintLinearGradient: 4>: 4939,
 <PaintFormat.PaintColrLayers: 1>: 6237,
 <PaintFormat.PaintSolid: 2>: 3604,
 <PaintFormat.PaintTransform: 12>: 21048,
 <PaintFormat.PaintComposite: 32>: 200,
 <PaintFormat.PaintTranslate: 14>: 248,
 <PaintFormat.PaintRadialGradient: 6>: 6275,
 <PaintFormat.PaintScale: 16>: 1015}
```

Most of the additional paints are PaintTransforms. Something to do with the new way in which nanoemoji implemented shape reuse (cf. https://github.com/googlefonts/nanoemoji/issues/302)

[...]

PR #319, opened 2021-08-08
Emit scale around center where appropriate

rsheeter:
Take advantage of PaintScaleAroundCenter where possible

### DIFF
--- a/src/nanoemoji/paint.py
+++ b/src/nanoemoji/paint.py
@@ -599,7 +599,7 @@ def is_transform(paint_or_format) -> bool:
 
 
 def _int16_safe(*values):
-    return all(v == int(v) and v <= 32767 and v >= -32768 for v in values)
+    return all(almost_equal(v, int(v)) and v <= 32767 and v >= -32768 for v in values)
 
 
 def _f2dot14_safe(*values):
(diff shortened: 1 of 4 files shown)


rsheeter:
Expected - but not verified in a real Noto compile - benefits outlined in https://github.com/googlefonts/nanoemoji/issues/316#issuecomment-894865790.

anthrotype:
one problem with replacing general PaintTransform with specialized PaintScaleAroundCenter is that the latter uses integers for the (cx, cy) center coordinates, so if you pass a float value then fontTools will round it to int as it compiles the PaintScaleAroundCenter.
E.g. your code converts `Affine2D(a=0.117, b=0, c=0, d=0.117, e=44.167, f=44.167)` to:

```python
cx = cy = 50.01925254813137
Affine2D.identity().translate(cx, cy).scale(0.117, 0.117).translate(-cx, -cy)
```
However when 50.01925254813137 becomes 50, the combination of translate, scale and counter-translate equals to `Affine2D(a=0.117, b=0.0, c=0.0, d=0.117, e=44.15, f=44.15)` , so there will be a ~= 0.017 translation delta in both x and y that is basically lost to rounding.

I guess we can deem that acceptable and live with it..

rsheeter:
> integers for the (cx, cy) center coordinates

Ah good catch, that's a bug in how I did this

rsheeter:
Added test + fix for the int'ness of cx, cy

anthrotype (approved):

PR #320, opened 2021-08-08
Tool to see breakdown of paints in a COLR v1 table

rsheeter:
Not the worlds most lovely code but I think it's helpful to have a shared copy. Traverse lifted from https://github.com/rsheeter/fonttools/blob/colrv1-varco/Lib/fontTools/ttLib/tables/otTables.py#L1272


tools/about_paint.py
```diff
@@ -0,0 +1,117 @@
...
+import sys
+
+def traverse_paint(colr, callback):
+    """Traverse and optionally rewrite Paint in a COLRv1 graph."""
+    if not callable(callback):
+        raise TypeError("callback must be callable")
+
+    visited = set()
+    frontier = []
+    # Glyph graphs start from either base glyphs or the layerlist
```
anthrotype, line 20: glyph graphs _only_ start from base glyphs, not from layer list. If any Paint in LayerList is not referenced by any base glyph (unlikely), it's as if it's not even there

tools/about_paint.py
```diff
@@ -0,0 +1,117 @@
...
+            _angle(asin, b),
+            _angle(lambda v: -asin(v), c),
+            _angle(acos, d)
+        }) == 1
+    )
+
+colr_builder = LayerListBuilder()
+
+def _tuple(paint):
+    return list_builder._paint_suble(paint)
```
anthrotype, line 72: typo?

```suggestion
    return list_builder._paint_tuple(paint)
```

anthrotype: `list_builder` name is undefined. I think the whole `_tuple(paint)` helper function is not used

tools/about_paint.py
rsheeter: It would be valid but pointless to have a stranded glyph in the layer list I think?

tools/about_paint.py
rsheeter: Quite right, copy/paste error

rsheeter: Fixed

tools/about_paint.py
anthrotype: indeed, but my point was simply that one normally should start traversing the COLR layer graphs from the root of a BaseGlyph record which is the only entry point for a renderer.

anthrotype: you already keep tally of the visited paints. for a normal well-behaved font (containing no dead paints in the LayerList) the end result of the traversal should be the same whether you include or not the top-level paints in the LayerList. I lean towards only including the root paints in the base glyph records.

rsheeter: I think here - unlike in a renderer - I do care about unreachable Paint since I just want to know what all is in the table?

rsheeter:
Amended to allow for in requirement on cx, cy

tools/about_paint.py
anthrotype: >  I just want to know what all is in the table

good point

tools/about_paint.py
```diff
@@ -51,6 +52,9 @@ def traverse_paint(colr, callback):
 def f2dot14_safe(*values):
     return all(value >= -2.0 and value < 2.0 for value in values)
 
+def int16_safe(*values):
+    return all(-32768 <= value < 32768 and int(value) == round(value, 3) for value in values)
```
anthrotype, line 56: in PR #319 (https://github.com/googlefonts/nanoemoji/pull/319/commits/8e80a190aa4e8870514645d1fe1f8309662196b9) you used `almost_equals(value, int(value))` instead of `== round(value, 3)`, should you do the same here?

tools/about_paint.py
```diff
@@ -92,17 +96,37 @@ def _about_paint(paint):
         transform = (tr.xx, tr.yx, tr.xy, tr.yy, tr.dx, tr.dy)
         # just scale and/or translate?
         if tr.xy == 0 and tr.yx == 0:
-            if f2dot14_safe(tr.xx, tr.yy):
-                if tr.dx == 0 and tr.dy == 0:
-                    count_by_type[paint.getFormatName() + "::scale_origin"] += 1
+            # Relationship between scale and translate leads to div 0?
+            if ((tr.xx == 1) != (tr.dx == 0)) or ((tr.yy == 1) != (tr.dy == 0)):
+                count_by_type[paint.getFormatName() + "::weird_scale"] += 1
```
anthrotype, line 101: there's nothing "weird" per se, it's just that you can't derive a PaintScaleAroundCenter with those

anthrotype (approved):

tools/about_paint.py
rsheeter: I think I'd go so far as to say we should have a shared function even

tools/about_paint.py
rsheeter: But why would someone do such a thing?! It's monstrous!

PR #325, opened 2021-08-12
fix transform for reused glyph with transformed gradients

### LINKED ISSUE #324: PaintGlyph reuse is broken when painted with transformed radial gradients
Take noto-emoji's [emoji_u1f60a.svg](https://github.com/googlefonts/noto-emoji/blob/main/svg/emoji_u1f60a.svg).  If you compile a COLRv1 font with it, and render in the latest Chrome Canary (Version 94.0.4605.0) which supports the latest COLRv1 spec changes, you'll see the right-hand side's red cheek disappears:

Screenshot 2021-08-12 at 20 41 54

The expected rendering should be instead:

Screenshot 2021-08-12 at 20 42 08

The problem seems to occur when a glyph reuse match is found (inside [`_migrate_paths_to_ufo_glyphs`](https://github.com/googlefonts/nanoemoji/blob/be4473f07d0baaa1102219be8b176e849c8f4e9f/src/nanoemoji/write_font.py#L271-L281) function), and the reused glyph gets wrapped in a PaintTransform (or one of the specialised variant), but the PaintGlyph.paint in turn is not a plain solid or gradient paint, but it is in turn one of the transform paints wrapping a gradient (we actually only do that for radial gradients currently, e.g. when gradientTransform squeezes their geometry to ellipses).
The result of concatenating the outer PaintTransform (from the reuse_result) with the inner transform makes the gradient look off.

The fix should be to concatenate the _inverse_ of the `reuse_result.transform` to the child transform wrapping the gradient, thus reverting its effect and restoring the coordinate space in which the gradient geometry was originally drawn.

I have a patch to that effect, i'll submit that soon.

anthrotype:
Fixes #324

### DIFF
--- a/src/nanoemoji/write_font.py
+++ b/src/nanoemoji/write_font.py
@@ -276,11 +276,17 @@ def _migrate_paths_to_ufo_glyphs(
             # TODO: when is it more compact to use a new transforming glyph?
             child_paint = paint.paint
             if is_transform(child_paint) and is_gradient(child_paint.paint):
-                child_paint = child_paint.paint.apply_transform(
-                    Affine2D.compose_ltr(
-                        (child_paint.gettransform(), reuse_result.transform.inverse())
-                    )
+                # We have a transformed gradient so we need to reverse the effect of the
+                # reuse_result.transform. First we try to apply the combined transform
+                # to the gradient's geometry; but this may overflow OT integer bounds,
+                # in which case we pass through gradient unscaled
+                transform = Affine2D.compose_ltr(
+                    (child_paint.gettransform(), reuse_result.transform.inverse())
                 )
+                try:
+                    child_paint = child_paint.paint.apply_transform(transform)
+                except OverflowError:
+                    child_paint = transformed(transform, child_paint.paint)
             return transformed(
                 reuse_result.transform,
                 PaintGlyph(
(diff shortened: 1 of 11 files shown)


tests/paint_test.py
```diff
@@ -192,3 +192,32 @@ def test_gettransform(input_point, paint, expected_point):
 )
 def test_transformed(transform, target, expected_result):
     assert transformed(transform, target) == expected_result
+
+
+@pytest.mark.parametrize(
+    "transform, expected",
+    [
+        (
+            Affine2D(2, 0, 0, 1, 0, 0),
```
rsheeter, line 201: Commenting the test cases would make this much more readable imo

anthrotype: I simply moved the existing test_decompose_uniform_transform from another test module (since I have moved the tested function to paint.py where it's used).
I'm not sure what I should write in the comment. Maybe if I split the fixture `expected` into two values `uniform_transform` and `remaining_transform` would be clearer?

src/nanoemoji/write_font.py
```diff
@@ -272,11 +274,18 @@ def _update_paint_glyph(paint):
         if reuse_result is not None:
             # TODO: use the most compact valid transform
```
anthrotype, line 275: I think this TODO is stale, since you have implemented the `transformed` function which is used immediately below. Shall I remove?

rsheeter: I think so, yes

Issue #334, opened 2021-09-06
Incorrect color line coordinates with shape reuse?


drott:
With a simplified traffic light glyph, the color line positions for the radial gradients seem off in the resulting Skia rendering.

The highlights on the red and green light are missing.

![shape_reuse](https://user-images.githubusercontent.com/1021784/132227216-aeb2171f-1bff-4de1-b9fe-a308df982963.png)

Simplified SVG

```
<?xml version="1.0" encoding="utf-8"?>
<!-- Generator: Adobe Illustrator 24.0.3, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
	 viewBox="0 0 128 128" style="enable-background:new 0 0 128 128;" xml:space="preserve">
  <radialGradient id="SVGID_4_" cx="55.969" cy="60.2974" r="13.562" gradientTransform="matrix(1 0 0 -1 3.9894 121.4094)" gradientUnits="userSpaceOnUse">
		<stop  offset="0.0139" style="stop-color:#FFEB3B"/>
		<stop  offset="0.6259" style="stop-color:#FCCD31"/>
		<stop  offset="1" style="stop-color:#FBC02D"/>
  </radialGradient>
  <circle style="fill:url(#SVGID_4_);" cx="64" cy="63.99" r="13.56"/>
  
  <radialGradient id="SVGID_5_" cx="55.824" cy="23.7944" r="13.562" gradientTransform="matrix(1 0 0 -1 3.9894 121.4094)" gradientUnits="userSpaceOnUse">
		<stop  offset="5.118100e-03" style="stop-color:#81C784"/>
		<stop  offset="0.2014" style="stop-color:#72BE76"/>
		<stop  offset="0.7194" style="stop-color:#50A854"/>
		<stop  offset="1" style="stop-color:#43A047"/>
  </radialGradient>
  <circle style="fill:url(#SVGID_5_);" cx="64" cy="100.93" r="13.56"/>

  <radialGradient id="SVGID_6_" cx="55.776" cy="97.9044" r="13.562" gradientTransform="matrix(1 0 0 -1 3.9894 121.4094)" gradientUnits="userSpaceOnUse">
	<stop  offset="0" style="stop-color:#E57373"/>
[...]

Issue #313, opened 2021-08-04
Generation of bad affines


rsheeter:
Observed with Noto and Twemoji, it seems there is a bug in how we produce transforms somewhere. For example:

In a color-fonts checkout, save the following in config/bad_affine.config:

```
family = "bad_affine-glyf_colr_1"
output_file = "bad_affine-glyf_colr_1.ttf"
color_format = "glyf_colr_1"

[axis.wght]
name = "Weight"
default = 400

[master.regular]
style_name = "Regular"
srcs = ["../font-srcs/twemoji/assets/svg/1f1e6-1f1f4.svg", "../font-srcs/twemoji/assets/svg/1f558.svg"]

[master.regular.position]
wght = 400
```

Try to build it:
```
nanoemoji --config bad_affine.config

I0803 20:11:39.007790 140004993345344 __init__.py:234] Building OpenType tables
Traceback (most recent call last):
  File "/usr/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "oss/nanoemoji/src/nanoemoji/write_font.py", line 640, in <module>
    app.run(main)
  File "oss/color-fonts/venv/lib/python3.9/site-packages/absl/app.py", line 312, in run
    _run_main(main, args)
  File "oss/color-fonts/venv/lib/python3.9/site-packages/absl/app.py", line 258, in _run_main
    sys.exit(main(argv))
  File "oss/nanoemoji/src/nanoemoji/write_font.py", line 634, in main
    ufo, ttfont = _generate_color_font(font_config, inputs)
  File "oss/nanoemoji/src/nanoemoji/write_font.py", line 600, in _generate_color_font
    ttfont = _make_ttfont(config, ufo, color_glyphs)
  File "oss/nanoemoji/src/nanoemoji/write_font.py", line 197, in _make_ttfont
    ttfont = ufo2ft.compileTTF(ufo, overlapsBackend="pathops")
  File "oss/color-fonts/venv/lib/python3.9/site-packages/ufo2ft/__init__.py", line 252, in compileTTF
    postProcessor = postProcessorClass(otf, ufo, glyphSet=glyphSet)
  File "oss/color-fonts/venv/lib/python3.9/site-packages/ufo2ft/postProcessor.py", line 50, in __init__
    self.otf = _reloadFont(otf)
[...]

PR #339, opened 2021-09-16
apply inverse of shape-reuse transform on gradient geometries

anthrotype:
When a shape get reused (wrapped by PaintTransform) and it's filled with a gradient, we must apply the inverse of the shape-reuse transform to the gradient's geometry to counteract its effect, as the gradient referred to the original coordinate system.

We were already doing something similar, but only for gradients that were in turn wrapped by a PaintTransform (#325). But @drott correctly pointed out that all the gradients after shape reuse need that treatment. 

Fixes https://github.com/googlefonts/nanoemoji/issues/334

After this patch is applied, Dominik's test SVG look like this 

Screenshot 2021-09-16 at 11 06 24


drott (approved):
Non owner LGTM, fix strategy seems reasonable to me and what I had assumed was going on. Thanks for the quick fix!

anthrotype:
@drott and I noticed that a `struct.error` with overflowing affines (similar to https://github.com/googlefonts/nanoemoji/issues/313 ) shows up after this PR when building the full noto-emoji.
The glyphs involved are the following:

```
fontTools.colorLib.errors.ColorLibError: Failed to build BaseGlyphList:
    g_1f6b4_1f3fb_200d_2642 => OverflowError: 174169.95561594606
    g_1f6b4_1f3fc_200d_2642 => OverflowError: 174169.95561594606
    g_1f6b4_1f3fd_200d_2642 => OverflowError: 174169.95561594606
    g_1f6b4_1f3fe_200d_2642 => OverflowError: 174169.95561594606
    g_1f6b4_1f3ff_200d_2642 => OverflowError: 174169.95561594606
    g_1f6b4_200d_2642 => OverflowError: 174169.95561594606 
    g_1f6b5 => OverflowError: 174128.0745902408
    g_1f6b5_1f3fb => OverflowError: 174128.0745902408
    g_1f6b5_1f3fb_200d_2640 => OverflowError: 174169.95561594606
    g_1f6b5_1f3fc => OverflowError: 174128.0745902408
    g_1f6b5_1f3fc_200d_2640 => OverflowError: 174169.95561594606
    g_1f6b5_1f3fd => OverflowError: 174128.0745902408
    g_1f6b5_1f3fd_200d_2640 => OverflowError: 174169.95561594606
    g_1f6b5_1f3fe => OverflowError: 174128.0745902408
    g_1f6b5_1f3fe_200d_2640 => OverflowError: 174169.95561594606
    g_1f6b5_1f3ff => OverflowError: 174128.0745902408
    g_1f6b5_1f3ff_200d_2640 => OverflowError: 174169.95561594606
    g_1f6b5_200d_2640 => OverflowError: 174169.95561594606
    g_1f916 => OverflowError: 243411.41639057896
    g_1f9bc => OverflowError: 141271.20406442852
    g_1f9f3 => OverflowError: 127548.34792178334
    g_1fa91 => OverflowError: 67216.62217602256
``` 

(I printed the names of the offending glyphs using [this wip fonttools branch](https://github.com/fonttools/fonttools/tree/colorLib-check-overflow))

If I build only some or all of those glyphs, then the build completes successfully, it's only when building the full noto-emoji that these struct.error occurs, which is kind of weird.

I think it's probably to do with the effect of the inverse() operation. Will continue to investigate.
I think we need to either:

1) skip reuse when this occurs
2) or break up huge transform into chains of non-overflowing PaintTransforms

rsheeter:
Typically this affine issue can be narrowed to a small set of svgs (ideally 2). If we can do that then perhaps try reversing the order (swap the names of files) and see if the result differs. My suspicion is that this relates to how ordering influences reuse and I need to enhance reuse to be a bit more resistant to this. My intuition is that prefering to reuse larger shapes to produce smaller shapes may help; the numbers don't blow up the way reusing small shapes to produce larger ones does.

anthrotype:
> prefering to reuse larger shapes to produce smaller shapes may help

Maybe, yes. In the meantime, in order to be able to finish the full noto-emoji build, I am disabling shape reuse when the applying inverse operation produces an affine that no longer fits the FIXED limits (https://github.com/googlefonts/nanoemoji/commit/01cd69a50a86920462269ae30285b45051c6da8d)

Also, I found a bug in picosvg's Affine2D.inverse() method (https://github.com/googlefonts/picosvg/pull/244), which is now fixed in v0.18.2. Thus make sure to update when testing the latest changes in this PR.

src/nanoemoji/write_font.py
```diff
@@ -273,26 +274,38 @@ def _update_paint_glyph(paint):
...
             # TODO: when is it more compact to use a new transforming glyph?
+            child_transform = Affine2D.identity()
             child_paint = paint.paint
-            if is_transform(child_paint) and is_gradient(child_paint.paint):
-                # We have a transformed gradient so we need to reverse the effect of the
+            if is_transform(child_paint):
+                child_transform = child_paint.gettransform()
+                child_paint = child_paint.paint
+
+            assert fixed_safe(*reuse_result.transform)
```
rsheeter, line 283: Provide a message, particularly anything that would inform user what glyph has the problem. I'm not sure if Python follows quite the same conventions but in other languages I'd say this should raise not assert

src/nanoemoji/write_font.py
```diff
@@ -273,26 +274,38 @@ def _update_paint_glyph(paint):
...
+            overflows = False
+            if is_gradient(child_paint):
```
rsheeter, line 285: What if there is a gradient that isn't the direct child?

I don't think we actually *DO* this so maybe a TODO would suffice.

tests/reused_shape_with_gradient_from_colr_1.svg
```diff
@@ -0,0 +1,24 @@
...
+      <stop offset="0" stop-color="#E57373"/>
+      <stop offset="0.197" stop-color="#E56564"/>
+      <stop offset="0.718" stop-color="#E54542"/>
+      <stop offset="1" stop-color="#E53935"/>
+    </radialGradient>
+  </defs>
+  <path fill="url(#g1)" d="M77.60000000000001,64 Q77.60000000000001,60.26666666666667 75.73333740234375,57.17333221435547 Q73.86666666666667,54.080000000000005 70.82666778564453,52.26667022705078 Q67.78666666666668,50.45333333333334 64.05333333333334,50.45333333333334 Q60.21333333333334,50.45333333333334 57.17333221435547,52.26667022705078 Q54.13333333333334,54.080000000000005 52.26667022705078,57.17333221435547 Q50.400000000000006,60.26666666666667 50.400000000000006,64 Q50.400000000000006,67.73333333333335 52.26667022705078,70.82666015625 Q54.13333333333334,73.92000000000002 57.17333221435547,75.73333740234375 Q60.21333333333334,77.54666666666668 64.05333333333334,77.54666666666668 Q67.78666666666668,77.54666666666668 70.82666778564453,75.73333740234375 Q73.86666666666667,73.92000000000002 75.73333740234375,70.82666015625 Q77.60000000000001,67.73333333333335 77.60000000000001,64 Z"/>
```
rsheeter, line 21: are we missing a rounding step somewhere?

rsheeter (approved):
Could nits noted but generally lg

src/nanoemoji/write_font.py
anthrotype: I deliberately made it an assertion instead of an exception because the shape reuse code already ensures that only fixed-safe transforms are used and skips those that would overflow, so I deem this an issue of internal code logic which should never be raised. I can add a comment to clarify that

tests/reused_shape_with_gradient_from_colr_1.svg
anthrotype: done

src/nanoemoji/write_font.py
anthrotype: thanks for noticing that. Indeed, we do not. The code only handles gradients that are direct child of PaintGlyph or child of child of PaintGlyph, via PaintTransform, which is the only subtree that we currently produce. I'll add the TODO

Issue #335, opened 2021-09-06
ClipBox too small with transformed reused shape (circus tent emoji example)


drott:
Simplifying the circus tent emoji, in the Skia rendering, the second yellow patch of the vertical tent wall goes missing. This path seems to be a mirrored version of the left side left patch of the vertical wall. 

Whether it gets shifted out of the viewport or whether something goes wrong with the color line coordinates, similar to issue #334 needs investigation.

![tent](https://user-images.githubusercontent.com/1021784/132232029-3d393fc3-624a-43a9-a4a1-4c12e3261afa.png)

Simplified SVG

```
<svg enable-background="new 0 0 128 128" viewBox="0 0 128 128"
	xmlns="http://www.w3.org/2000/svg">
	<linearGradient id="g" x1="32.5" x2="32.5" y1="64.167" y2="125.94" gradientUnits="userSpaceOnUse">
		<stop stop-color="#FDD835" offset="0"/>
		<stop stop-color="#F9A825" offset="1"/>
	</linearGradient>
	<path d="m20 64s12 24 8 60h16s4-44-4-60h-20z" fill="url(#g)"/>
	<linearGradient id="f" x1="95.5" x2="95.5" y1="64.167" y2="125.94" gradientUnits="userSpaceOnUse">
		<stop stop-color="#FDD835" offset="0"/>
		<stop stop-color="#F9A825" offset="1"/>
	</linearGradient>
	<path d="m108 64s-12 24-8 60h-16s-4-44 4-60h20z" fill="url(#f)"/>
</svg>
```

COLR

```
 <COLR>
    <Version value="1"/>
    <!-- BaseGlyphRecordCount=0 -->
    <!-- LayerRecordCount=0 -->
    <BaseGlyphList>
      <!-- BaseGlyphCount=1 -->
      <BaseGlyphPaintRecord index="0">
        <BaseGlyph value="u1F3AA"/>
        <Paint Format="1"><!-- PaintColrLayers -->
          <NumLayers value="2"/>
          <FirstLayerIndex value="0"/>
        </Paint>
      </BaseGlyphPaintRecord>
    </BaseGlyphList>
    <LayerList>
      <!-- LayerCount=2 -->
      <Paint index="0" Format="10"><!-- PaintGlyph -->
        <Paint Format="4"><!-- PaintLinearGradient -->
          <ColorLine>
            <Extend value="pad"/>
            <!-- StopCount=2 -->
            <ColorStop index="0">
[...]

PR #340, opened 2021-09-20
Fix bounding box incorrectly computed with reflection affine transform

anthrotype:
Fixes https://github.com/googlefonts/nanoemoji/issues/335

I first add a test to reproduce the bug, then follow up with the actual fix

### DIFF
--- a/src/nanoemoji/write_font.py
+++ b/src/nanoemoji/write_font.py
@@ -23,7 +23,7 @@ import csv
 import dataclasses
 import math
 from fontTools import ttLib
-from fontTools.misc.arrayTools import rectArea, unionRect
+from fontTools.misc.arrayTools import rectArea, normRect, unionRect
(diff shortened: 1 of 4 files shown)


anthrotype:
cf. failed CI test log reproducing the bug:

https://github.com/googlefonts/nanoemoji/pull/340/checks?check_run_id=3653210543#step:5:94

src/nanoemoji/write_font.py
```diff
@@ -23,7 +23,7 @@
 import dataclasses
 import math
 from fontTools import ttLib
-from fontTools.misc.arrayTools import rectArea, unionRect
+from fontTools.misc.arrayTools import rectArea, normRect, unionRect
```
rsheeter, line 26: aside: We should consider implementing these on the picosvg Rect, would be quite handy there

rsheeter (approved):
Good catch!

PR #342, opened 2021-09-21
Fix computing bounding box for rotated reused shapes

### LINKED ISSUE #341: incorrect clipbox for reused shapes that use rotation
nanoemoji is computing incorrect bounding boxes for reused shapes with a rotation affine transformation.
The problem is it first computes the bounds of the original (unrotated) shape, and then it applies the rotation to the bound rectangle, but that is wrong when rotation is involved (it works with basic scale or translation).
Nanoemoji should instead apply the transformation and then compute the bounds on the transformed shape itself.

image

anthrotype:
Fixes #341 

I'll add the (failing) test case first, then follow up with the fix

### DIFF
--- a/src/nanoemoji/write_font.py
+++ b/src/nanoemoji/write_font.py
@@ -26,6 +26,8 @@ from fontTools import ttLib
 from fontTools.misc.arrayTools import rectArea, normRect, unionRect
 from fontTools.misc.roundTools import otRound
 from fontTools.ttLib.tables import otTables as ot
+from fontTools.pens.boundsPen import ControlBoundsPen
+from fontTools.pens.transformPen import TransformPen
 from itertools import chain
 from lxml import etree  # pytype: disable=import-error
 from nanoemoji import codepoints, config
@@ -461,6 +463,17 @@ def _quantize_bounding_rect(
     )
 
 
+def _transformed_glyph_bounds(
+    ufo: ufoLib2.Font, glyph_name: str, transform: Affine2D
+) -> Optional[Tuple[float, float, float, float]]:
+    glyph = ufo[glyph_name]
+    pen = bounds_pen = ControlBoundsPen(ufo)
+    if transform != Affine2D.identity():
+        pen = TransformPen(bounds_pen, transform)
+    glyph.draw(pen)
+    return bounds_pen.bounds
+
+
 def _bounds(
     color_glyph: ColorGlyph, quantize_factor: int = 1
 ) -> Optional[Tuple[int, int, int, int]]:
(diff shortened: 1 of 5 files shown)


anthrotype:
https://github.com/googlefonts/nanoemoji/pull/342/checks?check_run_id=3662370188#step:5:103 => you can see how the computed bounds here is wrong, it gets flattened to a line making the rotated shape disappear

anthrotype:
**BEFORE**

image

**AFTER**

image

anthrotype:
I force-pushed the first commit (with the failing test case only) to fix up an inline comment, but I just realised that the CI log for the old commit gets lost as well.. Anyway, you got the idea.

src/nanoemoji/write_font.py
```diff
@@ -461,6 +463,17 @@ def _quantize_bounding_rect(
...
 
+def _transformed_glyph_bounds(
+    ufo: ufoLib2.Font, glyph_name: str, transform: Affine2D
+) -> Optional[Tuple[float, float, float, float]]:
+    glyph = ufo[glyph_name]
+    pen = bounds_pen = ControlBoundsPen(ufo)
+    if transform != Affine2D.identity():
+        pen = TransformPen(bounds_pen, transform)
+    glyph.draw(pen)
+    return bounds_pen.bounds
```
rsheeter, line 474: Much better.

src/nanoemoji/write_font.py
```diff
@@ -461,6 +463,17 @@ def _quantize_bounding_rect(
     )
 
 
+def _transformed_glyph_bounds(
+    ufo: ufoLib2.Font, glyph_name: str, transform: Affine2D
+) -> Optional[Tuple[float, float, float, float]]:
+    glyph = ufo[glyph_name]
+    pen = bounds_pen = ControlBoundsPen(ufo)
+    if transform != Affine2D.identity():
```
rsheeter, line 471: Hm. Almost equals possibly?

tests/write_font_test.py
```diff
@@ -329,15 +337,18 @@ def test_ufo_color_base_glyph_bounds(svgs, config_overrides, expected):
     config, glyph_inputs = test_helper.color_font_config(config_overrides, svgs)
     ufo, _ = write_font._generate_color_font(config, glyph_inputs)
 
-    base_glyph_name = "e000"
-    assert len(ufo[base_glyph_name]) == 0
+    base_glyph_names = [f"e{str(i).zfill(3)}" for i in range(len(svgs))]
```
rsheeter, line 340: (apparently I can't comment unmodified lines)

I think maybe you should rename expected, it wasn't a very good name to begin with and now it's even less clear. expected_clip_boxes perhaps?

src/nanoemoji/write_font.py
anthrotype: Done https://github.com/googlefonts/nanoemoji/commit/a365e618a8595d261c6c6159b4384632582a463e

tests/write_font_test.py
anthrotype: Done https://github.com/googlefonts/nanoemoji/commit/a365e618a8595d261c6c6159b4384632582a463e

PR #343, opened 2021-09-21
fix a couple of bugs in tools/colr2svg.py script 

anthrotype:
with this, I can fully convert all the glyphs from noto_emoji-glyf_colr_1.ttf to SVG files using the tools/colr2svg.py script.
See commit message for details


narinishi:
Hello @anthrotype, it appears the colr2svg tool has been broken for a while, might you be able to fix it? Thanks for your consideration.

anthrotype:
what error do you get? Do you mind filing a new issue with a way to reproduce?

Issue #337, opened 2021-09-15
[OT-SVG] picosvg shape reuse not working when reused paths have different fill  


anthrotype:
Take this simple SVG containing three rects of different colors:

```xml
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">
  <rect fill="yellow" x="40" y="10" width="20" height="20"/>
  <rect fill="red" x="40" y="40" width="20" height="20" />
  <rect fill="blue" x="40" y="70" width="20" height="20" />
</svg>
```
![image](https://user-images.githubusercontent.com/6939968/133472167-bd9a2b54-83d7-40ee-ae67-5b9b513259df.png)

When compiling an OT-SVG font from it with the command `nanoemoji --color_format=picosvg --pretty_print /tmp/e002.svg`, the resulting `SVG  ` table contains the following:

```xml
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1">
  <g id="glyph2" transform="matrix(12 0 0 12 37.5 -950)">
    <path d="M40,10 L60,10 L60,30 L40,30 L40,10 Z" fill="yellow" id="e002.0"/>
    <use xlink:href="#e002.0" y="30" fill="red"/>
    <use xlink:href="#e002.0" y="60" fill="blue"/>
  </g>
</svg>
```

All three rects render yellow, using the same fill as the first element from which they inherit the path.d.
![image](https://user-images.githubusercontent.com/6939968/133472297-505c297a-c031-4b6d-bf95-77db08cd0ac9.png)

The `fill` attribute on the `<use>` element is ignored because the referenced path already contains a `fill`, so the attribute does not "cascade" down, the reused path's fill takes priority over the `fill` defined onto the `<use>` element that references it.
Basically attributes on `<use>` do not override those explicitly defined on the referenced element, they can only be inherited when the latter doesn't define them.

We could fix this by moving the reused path (without its fill, only with the `d` attribute) to the `<defs>` and replace all occurences (including the first one) with `<use>` elements that each define their own distinct `fill`, e.g.:

```xml
<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1">
  <defs>
    <path d="M40,10 L60,10 L60,30 L40,30 L40,10 Z" id="e002.0"/>
  </defs>
[...]

PR #345, opened 2021-10-02
Fix bad assumption that <use> can override target fill

rsheeter:
Test to expose and prove fix works

Fixes https://github.com/googlefonts/nanoemoji/issues/337


src/nanoemoji/svg.py
```diff
@@ -340,7 +355,57 @@ def _apply_paint(
...
+    svg_defs = svg.xpath_one("//svg:defs")
+
+    if reused_el in svg_defs:
+        return  # nop
+
+    svg_use = etree.Element("use", nsmap=svg.svg_root.nsmap)
+    svg_use.attrib[_XLINK_HREF_ATTR_NAME] = f"#{reuse_result.glyph_name}"
+    # if reused_el hasn't been given a parent yet just let the <use> replace it
+    # otherwise move it from current to new parent
+    if reused_el.getparent() is None:
```
anthrotype, line 380: when does reused_el have no parent? I don't think _migrate_to_defs ever meets a parent-less reused_el.
IIUC, at a first occurrence of a paint glyph, when by definition there is no (not yet) reuse, we always append the corresponding element to the parent element (svg root or g), thus by the time a reuse is found, that reused element already got a parent.
If that's the case, I think here we should assert/assume that reused_el already has a parent.

anthrotype (approved):
Looks good, thank you!

src/nanoemoji/svg.py
rsheeter: Hm. It was definitely possible at some point but now I'm failing to come up with when that would be. I think I'm going to file a new ticket for that and leave it for this one.

rsheeter: https://github.com/googlefonts/nanoemoji/issues/346

PR #349, opened 2021-10-11
Compute reuse on the svg path; seems to increase odds of success

rsheeter:
DO NOT SUBMIT; needs a test that previously didn't find reuse and now does. Tested with https://github.com/googlefonts/picosvg/pull/245 applied.

Example:

```
nanoemoji ~/oss/color-fonts/font-srcs/noto-emoji/svg/emoji_u1f610.svg

# before [more glyphs / less COLR], 1556 bytes
tag   size   pct
head     54    3 
hhea     36    2 
maxp     32    2 
OS/2     96    6 
hmtx     24    2 
cmap    108    7 
loca     22    1 
glyf    402   26 10 glyphs
name    300   19 
post     32    2 
COLR    197   13 7 glyphs
CPAL     38    2

# after [fewer glyphs / more COLR], 1508 bytes
tag   size   pct
head     54    4 
hhea     36    2 
maxp     32    2 
OS/2     96    6 
hmtx     20    1 
cmap    108    7 
loca     18    1 
glyf    304   20 8 glyphs
name    300   20 
post     32    2 
COLR    259   17 5 glyphs
CPAL     38    3
```

### DIFF
--- a/src/nanoemoji/write_font.py
+++ b/src/nanoemoji/write_font.py
@@ -269,12 +269,22 @@ def _migrate_paths_to_ufo_glyphs(
             return paint
 
         assert paint.glyph.startswith("M"), f"{paint.glyph} doesn't look like a path"
+        path_in_svg_space = paint.glyph
         path_in_font_space = (
-            SVGPath(d=paint.glyph).apply_transform(svg_units_to_font_units).d
+            SVGPath(d=path_in_svg_space).apply_transform(svg_units_to_font_units).d
         )
 
-        reuse_result = glyph_cache.try_reuse(path_in_font_space)
+        print("REUSE IN SVG SPACE")
+        print("  ", path_in_svg_space)
+        print("  ", path_in_font_space)
+        # The odds of finding reuse in svg space seems to be much higher. Not completely sure why.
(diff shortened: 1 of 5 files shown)


src/nanoemoji/write_font.py
```diff
@@ -269,12 +269,28 @@ def _update_paint_glyph(paint):
...
         path_in_font_space = (
-            SVGPath(d=paint.glyph).apply_transform(svg_units_to_font_units).d
+            SVGPath(d=path_in_svg_space).apply_transform(svg_units_to_font_units).d
         )
 
-        reuse_result = glyph_cache.try_reuse(path_in_font_space)
+        print("REUSE IN SVG SPACE")
+        print("  ", path_in_svg_space)
+        print("  ", path_in_font_space)
+        # The odds of finding reuse in svg space seems to be much higher. Not completely sure why.
```
anthrotype, line 280: maybe because the operations involved in mapping from svg to font space introduce some floating-point loss of precision?

rsheeter: Plausible but not yet proven. I thought floats durable enough to survive the type of mangling we're doing but maybe that's just ... wrong.

rsheeter:
I prefer the approach in #410; closing this

PR #359, opened 2021-10-31
Suport writing a font with COLRv1 *and* ot-svg

### LINKED ISSUE #260: Ability to support multiple color tables
We have use cases where it would be helpful to have multiple color formats, such as COLR and SVG. Consider making this possible.

rsheeter:
Avoid mutating the original copy of color_glyphs. This makes gluing an svg table onto a COLRv1 font kind of Just Work.

Fixes #260

### DIFF
--- a/src/nanoemoji/colors.py
+++ b/src/nanoemoji/colors.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import collections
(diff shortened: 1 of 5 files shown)


anthrotype (approved):
LGTM!

PR #360, opened 2021-11-08
Don't add empty glyphs without painted layers to COLR or SVG table

anthrotype:
as found out in https://github.com/fonttools/fonttools/pull/2441/files#r744692672

### DIFF
--- a/tests/nanoemoji_test.py
+++ b/tests/nanoemoji_test.py
@@ -178,3 +178,26 @@ def test_glyphmap_games():
     assert_expected_ttx(
         svgs, font, "glyphmap_games.ttx", include_tables=["GlyphOrder", "cmap"]
     )
+
+
+def test_omit_empty_color_glyphs():
+    svgs = [
+        "emoji_u200c.svg",  # whitespace glyph, contains no paths
+        "emoji_u42.svg",
+    ]
+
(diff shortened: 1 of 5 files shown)


anthrotype:
oops. I had forgotten to add a test file, should work now

tests/nanoemoji_test.py
```diff
@@ -178,3 +178,26 @@ def test_glyphmap_games():
...
+        )
+    )
+
+    font = TTFont(tmp_dir / "Font.ttf")
+
+    colr = font["COLR"].table
+    assert len(colr.BaseGlyphList.BaseGlyphPaintRecord) == 1
+
+    svg = font["SVG "]
+    assert len(svg.docList) == 1
```
rsheeter, line 203: Optional. Not entirely convinced by the existence of just 1 ... shouldn't we confirm it appears to be the right one? - I think maybe assert_expected_ttx might be worthwhile instead of just len 1

rsheeter (approved):

tests/nanoemoji_test.py
anthrotype: yeah, in fact I did that initially, and found out about that other bug #361. I'll merge the latter first, then add assert_expected_ttx in here

PR #361, opened 2021-11-08
be more careful when reordering SVG color glyphs

### LINKED ISSUE #260: Ability to support multiple color tables

anthrotype:
don't simply assume they are always all at the end of the glyph order. For mixed COLR+SVG fonts, there may be additional layer glyphs after the block of base color glyphs.

Note how the expected test .ttx file had the wrong glyph index for the SVG color glyph: it was pointing to glyph 3 "e000.0" which is in fact one of the COLR layer glyphs, instead of glyph 2 "e000", the correct base color glyph.
 
Fixes #260

### DIFF
--- a/src/nanoemoji/svg.py
+++ b/src/nanoemoji/svg.py
@@ -45,6 +45,7 @@ from picosvg.svg_transform import Affine2D
 from picosvg.svg_types import SVGPath
 from typing import (
     cast,
+    Mapping,
     MutableMapping,
     NamedTuple,
     Optional,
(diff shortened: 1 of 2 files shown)


src/nanoemoji/svg.py
```diff
@@ -557,27 +558,34 @@ def _ensure_groups_grouped_in_glyph_order(
...
+    # while keeping the current names: we can't change order and rename at the same time
     # or else tables that contain mappings keyed by glyph name would blow up.
     # Thus, we need to match the old and current names by their position in the
-    # font's current glyph order: i.e. we assume all color glyphs are placed at the
-    # END of the glyph order.
+    # font's current glyph order: i.e. we assume all color glyphs are placed in a
+    # continuous block starting at the first color glyph.
     current_glyph_order = ttfont.getGlyphOrder()
-    current_color_glyph_names = current_glyph_order[-len(color_glyphs) :]
+    first_color_gid = min(g.glyph_id for g in color_glyphs.values())
```
rsheeter, line 568: Optional, you could do this in a reduce and get back a tuple (min, max). The slice below could then go from [min:max+1]

rsheeter:
Ew, good catch

rsheeter (approved):

src/nanoemoji/svg.py
anthrotype: simply taking min and max is much faster than doing a reduce in single pass

PR #362, opened 2021-11-09
use ninjia @*.rsp file in write_glyphmap.py

anthrotype:
otherwise I get 'ninja: fatal: posix_spawn: Argument list too long' on Linux

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,7 +15,9 @@ jobs:
     - name: Set up Python 3.9
       uses: actions/setup-python@v2
       with:
-        python-version: 3.9
+        # typed_ast (used by pytype) has a problem when run under 3.9.8, pin to 3.9.7
+        # until resolved: https://github.com/python/typed_ast/issues/169
+        python-version: "3.9.7"
     - name: Install tox
       run: pip install tox
     - name: Run style and typing checks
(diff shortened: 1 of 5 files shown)


anthrotype:
no idea what's going on on windows.. I see things like:

```
FileNotFoundError: [Errno 2] No such file or directory: 'picosvgregularclipped61.svg'
```

maybe pathlib is not correctly joining the subpaths using windows-specific backslashes? No idea how to test this without a windows machine at hand..

rsheeter (approved):

anthrotype:
the problem seems to be shlex.split function, used to split the content of the .rsp ninja file into back into a list of argv. Since it defaults to posix=True and the backslash is an escape character, a path like "picosvg\regular\61.svg" was getting eaten up on Windoes. At least setting posix=False (when os.name != "posix", e.g. when running on Windows) seems to make our tests pass.

But.. now the lint job is failing for an unrelated reason :(
pytype depends on typed_ast, and the latter seems to have some issues on Python 3.9
https://github.com/python/typed_ast/issues/169

Not sure what to do.. Maybe I'll try to temporarily pin pytype to an earlier version. Or try to run on an earlier python vesion.. Why is it so hard... :(

anthrotype:
lint job passes with 3.9.7. I filed an issue upstream https://github.com/google/pytype/issues/1044

PR #366, opened 2022-01-13
workaround for Skia bug whereby linear gradient's P2 has no effect

anthrotype:
because of a regression in Skia, skewed linear gradients (i.e. where P2 is not perpendicular to P1-P0 vector) don't render correctly in Chrome. The position of rotation point P2 has no effect and the gradient never gets rotated.

See upstream issues:
https://skia-review.googlesource.com/c/skia/+/494676
https://bugs.chromium.org/p/chromium/issues/detail?id=1287162

As workaround, here we normalize the linear gradients so that they will look the same whether or not P2 is taken into account, by setting P1 to its projection onto the line perpendicular to the rotation vector.
We can remove this once the bug is fixed in Chrome.

### DIFF
--- a/src/nanoemoji/color_glyph.py
+++ b/src/nanoemoji/color_glyph.py
@@ -137,9 +137,17 @@ def _parse_linear_gradient(
         config, grad_el, shape_bbox, view_box, glyph_width
     )
 
-    return PaintLinearGradient(  # pytype: disable=wrong-arg-types
+    paint = PaintLinearGradient(  # pytype: disable=wrong-arg-types
         p0=p0, p1=p1, p2=p2, **common_args
     ).apply_transform(transform)
+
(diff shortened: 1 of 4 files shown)


anthrotype:
Let's hold this one for a bit, Dominik's fix has already been merged in Skia, and should appear in Chrome Canary in the coming days. If the fix is backported to 98 branch we may no longer need to add the workaround.

rsheeter (approved):
LGTM if it proves necessary, hopefully it doesn't

anthrotype:
It appears the latest Chrome Canary Version 99.0.4838.3 contains the fix. Fingers crossed it'll be backported to m98

anthrotype:
the bugfix was merged to Skia's chrome/m98 branch so we don't need this any more

PR #370, opened 2022-01-17
Bitmap support, part 1: just enough sbix and cbdt to produce a trivial font that renders

### LINKED ISSUE #88: Support all 4 OpenType color formats
Migrating @davelab6 request in https://github.com/googlefonts/noto-emoji/issues/259 to here because if we ever support all formats it'll be done here.

Format support moved to #370

rsheeter:
Testing production of bitmap tables.

Build on top of https://github.com/googlefonts/nanoemoji/tree/resvg.

I can specify `bitmapGlyphDataFormat="extfile"` in `test_helper.py` to avoid testing if resvg produces identical bitmaps on all platforms but it doesn't work for sbix. It would be very helpful if it did, @anthrotype wdyt? EDIT: for now I manually wiped out the sbix images.

Fixes #88


src/nanoemoji/bitmap_tables.py
```diff
@@ -0,0 +1,269 @@
...
+        print("line_ascent", metrics.line_ascent)
+        print("metrics.x_offset", metrics.x_offset)
+        print("metrics.y_offset", metrics.y_offset)
+
+        return metrics
+
+
+# https://github.com/googlefonts/noto-emoji/blob/9a5261d871451f9b5183c93483cbd68ed916b1e9/third_party/color_emoji/emoji_builder.py#L53
+def _ppem(config: FontConfig, advance: int) -> int:
+    return int(config.bitmap_resolution * config.upem / advance)
```
anthrotype, line 88: if I use bitmap_advance, I obtain the same ppem value as emoji_builder.py, and we don't need to add that `+ 8` hack further below (because 136 == 128 + 8):
https://github.com/googlefonts/noto-emoji/blob/9a5261d871451f9b5183c93483cbd68ed916b1e9/third_party/color_emoji/emoji_builder.py#L58

```suggestion
    return round(config.bitmap_advance * config.upem / advance)
```

Also note how the `div()` method used in emoji_builder.py does `round()` instead of `int()` (which truncates); we should do the same here to match exactly its output.

Getting the ppem right is crucial becase that's what determines the relatiove size of the bitmap

src/nanoemoji/bitmap_tables.py
```diff
@@ -0,0 +1,269 @@
...
+
+    sbix = ttLib.newTable("sbix")
+    ttfont[sbix.tableTag] = sbix
+
+    sbix.ppem = _ppem(config, _advance(ttfont, color_glyphs))
+    # TODO what goes here
+    sbix.ppi = 96
+
+    strike = SbixStrike()
+    strike.ppem = sbix.ppem + 8  # TODO magic #
```
anthrotype, line 154: I'm puzzled by the + 8. Also, why sbix.ppem would be different from SbixStrike.ppem?

BTW, I am getting weird rendering in Safari Mac, the bitmaps look too small and offset towards the top:

image

In Chrome Mac, instead I just see AdobeBlank fallback when rendering sbix... :(

src/nanoemoji/config.py
```diff
@@ -103,6 +103,14 @@
     None,
     "A program that takes a list of filenames and outputs a file csv whose rows contain filename, codepoint(s), glyph name.",
 )
+flags.DEFINE_integer(
+    "bitmap_resolution", None, "Resolution of bitmap in pixels. Always square for now."
+)
+flags.DEFINE_integer(
+    "bitmap_advance",
+    None,
+    "Advance for bitmap, in pixels. Typically >= bitmap_resolution.",
```
anthrotype, line 112: we don't need to add this option. We can derive the bitmap advance width from the hmtx width (which can be configured already). 

(ascender - descender) : width == bitmap_resolution : bitmap_width

bitmap_width = bitmap_resolution * width / (ascender - descender)

For our current defaults, this means:

bitmap_width = 128 * 1275 / 1200 == 136

src/nanoemoji/bitmap_tables.py
```diff
@@ -0,0 +1,269 @@
...
+def make_sbix_table(
+    config: FontConfig,
+    ttfont: ttLib.TTFont,
+    color_glyphs: Sequence[ColorGlyph],
+):
+
+    sbix = ttLib.newTable("sbix")
+    ttfont[sbix.tableTag] = sbix
+
+    sbix.ppem = _ppem(config, _advance(ttfont, color_glyphs))
```
anthrotype, line 149: I think this line is wrong, only the SbixStrike has ppem attribute, not the sbix table as a whole

src/nanoemoji/bitmap_tables.py
```diff
@@ -0,0 +1,269 @@
...
+    # TODO what goes here
+    sbix.ppi = 96
```
anthrotype, line 151: also this `sbix.ppi` doesn't exist, it's only on the SbixStrike, not the sbix table

anthrotype: actually, the OT spec calls the field `ppi` (density of pixels per inch) wheres the fontTools SbixStrike class names it (incorrectly) `resolution`.

The two terms are often confused. See:
https://en.wikipedia.org/wiki/Image_resolution
https://en.wikipedia.org/wiki/Pixel_density

> Note that pixel density is not the same as resolution, where the former describes the amount of detail on a physical surface or device, the latter describes the amount of pixel information regardless of its scale

src/nanoemoji/bitmap_tables.py
```diff
@@ -0,0 +1,269 @@
...
+
+    strike = SbixStrike()
+    strike.ppem = sbix.ppem + 8  # TODO magic #
+    strike.resolution = config.bitmap_resolution
```
anthrotype, line 155: Apple Color Emoji has 72 for this field, we should do the same.

With the patch below, I can finally see the correct size/placement in Safari Mac for the sbix font:

```diff
diff --git a/src/nanoemoji/bitmap_tables.py b/src/nanoemoji/bitmap_tables.py
index 61b8353..52d9e00 100644
--- a/src/nanoemoji/bitmap_tables.py
+++ b/src/nanoemoji/bitmap_tables.py
@@ -85,7 +85,7 @@ class BitmapMetrics(NamedTuple):
 
 # https://github.com/googlefonts/noto-emoji/blob/9a5261d871451f9b5183c93483cbd68ed916b1e9/third_party/color_emoji/emoji_builder.py#L53
 def _ppem(config: FontConfig, advance: int) -> int:
-    return int(config.bitmap_resolution * config.upem / advance)
+    return int(config.bitmap_advance * config.upem / advance)
 
 
 def _advance(ttfont: ttLib.TTFont, color_glyphs: Sequence[ColorGlyph]) -> int:
@@ -146,13 +146,12 @@ def make_sbix_table(
     sbix = ttLib.newTable("sbix")
     ttfont[sbix.tableTag] = sbix
 
-    sbix.ppem = _ppem(config, _advance(ttfont, color_glyphs))
+    ppem = _ppem(config, _advance(ttfont, color_glyphs))
     # TODO what goes here
-    sbix.ppi = 96
 
     strike = SbixStrike()
-    strike.ppem = sbix.ppem + 8  # TODO magic #
-    strike.resolution = config.bitmap_resolution
+    strike.ppem = ppem
+    strike.resolution = 72  # pixels per inch
     sbix.strikes[strike.ppem] = strike
 
     metrics = BitmapMetrics.create(config, strike.ppem)
@@ -167,7 +166,7 @@ def make_sbix_table(
             glyphName=glyph_name,
             imageData=image_data,
             originOffsetX=metrics.x_offset,
-            originOffsetY=metrics.line_height - metrics.y_offset,
+            originOffsetY=metrics.line_ascent - metrics.line_height,
         )
         strike.glyphs[glyph_name] = glyph
 
@@ -189,7 +188,7 @@ def make_cbdt_table(
     ), "Below assumes color gyphs gids are consecutive"
 
     advance = _advance(ttfont, color_glyphs)
-    ppem = _ppem(config, advance) + 8
+    ppem = _ppem(config, advance)
 
     cbdt = ttLib.newTable("CBDT")
     ttfont[cbdt.tableTag] = cbdt
```

image

src/nanoemoji/bitmap_tables.py
```diff
@@ -0,0 +1,269 @@
...
+        lambda a, c: (min(a[0], c.glyph_id), max(a[1], c.glyph_id)),
+        color_glyphs,
+        (sys.maxsize, -1),
+    )
+    assert max_gid - min_gid + 1 == len(
+        color_glyphs
+    ), "Below assumes color gyphs gids are consecutive"
+
+    advance = _advance(ttfont, color_glyphs)
+    ppem = _ppem(config, advance) + 8
```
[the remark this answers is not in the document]
anthrotype: Once we refactor the code that computes the ppem from the bitmap advance, head.unitsPerEm and hmtx.width, we can get rid of this +8 magic number.

src/nanoemoji/bitmap_tables.py
```diff
@@ -0,0 +1,269 @@
...
+                int((config.bitmap_advance - config.bitmap_resolution) / 2), 0
+            ),
+            y_offset=int(
+                round(line_ascent - 0.5 * (line_height - config.bitmap_resolution))
+            ),
+            line_height=line_height,
+            line_ascent=int(line_ascent),
+        )
+
+        print("line_height", metrics.line_height)
```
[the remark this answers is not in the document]
rsheeter: Oops, deleting

src/nanoemoji/bitmap_tables.py
rsheeter: Ha good catch. My kingdom for a complier :D

src/nanoemoji/config.py
rsheeter: Oh nice, good point. Fixed!

src/nanoemoji/bitmap_tables.py
rsheeter: Updated so but ofc ptal and confirm it looks sane

src/nanoemoji/config.py
rsheeter: Done

src/nanoemoji/bitmap_tables.py
rsheeter: ...I fear you are right

src/nanoemoji/write_font.py
```diff
@@ -108,6 +108,17 @@ class ColorGenerator(NamedTuple):
     font_ext: str  # extension for font binary, .ttf or .otf
 
 
+_OUTLINE_GENERATORS = {
```
anthrotype, line 111: this seems unused?

src/nanoemoji/write_font.py
```diff
@@ -108,6 +108,17 @@ class ColorGenerator(NamedTuple):
...
+    "none": lambda *_: None,
+    "glyf": lambda *_: None,
+    "cff": lambda *_: None,
+    "cff2": lambda *_: None,
+}
+# assert _OUTLINE_GENERATORS.keys() == set(config._OUTLINE_FORMATS)
+
+_COLOR_GENERATORS = {}
```
anthrotype, line 119: this also is unused

tests/rect_glyf_colr_1_and_picosvgz_and_sbix.ttx
```diff
@@ -0,0 +1,142 @@
...
+    <version value="0"/>
+    <numPaletteEntries value="1"/>
+    <palette index="0">
+      <color index="0" value="#0000FFFF"/>
+    </palette>
+  </CPAL>
+
+  <SVG>
+    <svgDoc endGlyphID="2" startGlyphID="2">
+      &lt;svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1"&gt;
```
anthrotype, line 116: why is this text escaped? should it not appear inside a CDATA block?

src/nanoemoji/bitmap_tables.py
```diff
@@ -0,0 +1,265 @@
...
+    line_height: int
+    line_ascent: int
+
+    @classmethod
+    def create(cls, config: FontConfig, ppem: int) -> "BitmapMetrics":
+        # https://github.com/googlefonts/noto-emoji/blob/9a5261d871451f9b5183c93483cbd68ed916b1e9/third_party/color_emoji/emoji_builder.py#L109
+        ascent = config.ascender
+        descent = -config.descender
+
+        line_height = int((ascent + descent) * ppem / float(config.upem))
```
anthrotype, line 64: the `/` operator always does float division (`//` is for integer division), you don't need to turn one of the opearands to float(). Also I believe it's better to`round` instead of truncating with `int`.

```suggestion
        line_height = round(ascent + descent) * ppem / config.upem
```

src/nanoemoji/bitmap_tables.py
```diff
@@ -0,0 +1,265 @@
...
+        line_ascent = ascent * ppem / float(config.upem)
+
+        metrics = BitmapMetrics(
+            x_offset=max(
+                int((_width_in_pixels(config) - config.bitmap_resolution) / 2), 0
```
anthrotype, line 69: you seem to sytematically prefer `int()` over `round()` -- curious what's the rationale?

src/nanoemoji/bitmap_tables.py
```diff
@@ -0,0 +1,265 @@
...
+
+def _width_in_pixels(config: FontConfig) -> int:
+    return int(
+        config.bitmap_resolution * config.width / (config.ascender - config.descender)
+    )
+
+
+# https://github.com/googlefonts/noto-emoji/blob/9a5261d871451f9b5183c93483cbd68ed916b1e9/third_party/color_emoji/emoji_builder.py#L53
+def _ppem(config: FontConfig, advance: int) -> int:
+    return int(_width_in_pixels(config) * config.upem / advance)
```
anthrotype, line 89: `_width_in_pixels` already rounds (truncates actually), then you multiply by a float and round (truncate) again -- I see this as a risk of losing precision, how about only rounding once at the end?

src/nanoemoji/bitmap_tables.py
```diff
@@ -0,0 +1,265 @@
...
+        # TODO: if we've seen these bytes before set graphicType "dupe", referenceGlyphName <name of glyph>
+        image_data = _image_bytes(color_glyph)
+
+        glyph_name = ttfont.getGlyphName(color_glyph.glyph_id)
+        glyph = SbixGlyph(
+            graphicType="png",
+            glyphName=glyph_name,
+            imageData=image_data,
+            originOffsetX=metrics.x_offset,
+            originOffsetY=metrics.line_height - metrics.y_offset,
```
anthrotype, line 169: I get weird vertical positioning in Safari Mac with your originOffsetY code:

Screenshot 2022-01-25 at 11 47 09

If I flip it like so, I get the expected placement:

```suggestion
            originOffsetY=metrics.y_offset - metrics.line_height,
```

Screenshot 2022-01-25 at 11 51 23

Also using line_ascent intead of y_offset looks ok (the two ought to be equal because we stretch to fit the vertical distance and only center horizontally -- actually if y_offset and line_ascent don't match exactly I'd count that as a bug):
```suggestion
            originOffsetY=metrics.line_ascent - metrics.line_height,
```

anthrotype: uh-oh... I think we found a bug in sbix Chrome implementation..
Chrome on Linux (first screenshot below) doesn't match Safari on Mac's treatment of originOffsetY (second screenshot).

Screenshot 2022-01-25 at 12 18 05

Screenshot 2022-01-25 at 11 47 09

This is using your current code for `originOffsetY=metrics.line_height - metrics.y_offset`.
If I invert that to `metrics.y_offset - metrics.line_height`, I fix Safari Mac but break Chrome Linux!
I attached the test font and static html page here:
[Archive.zip](https://github.com/googlefonts/nanoemoji/files/7933523/Archive.zip)

~~~

Actually, we maybe found two bugs: unlike Chrome on Linux, the very same sbix font and html page renders all blank in Chrome **on Mac**!

![image](https://user-images.githubusercontent.com/6939968/150976786-7d38704b-fc2f-4c86-9d77-9ffdffbcf970.png)

anthrotype: as for the second bug (sbix test font not displaying in Chrome for Mac), it's because nanoemoji is writing empty glyphs in glyf table -- ever since ClipBox were added to COLRv1 it no longer emits the two invisible glyf contour points that delimit the bounding box of the color glyph. Looks like Safari Mac doesn't care if the glyphs are empty and displays the color glyphs nonetheless, but _Chrome Mac won't display the sbix color glyphs unless the respective b/w glyf glyphs contain an outline whose bbox is enough to contain the color glyph_. So for sbix at least we need those back.

anthrotype: The sbix opentype spec has for `originOffsetY` (emphasis mine)

> The vertical (y-axis) position of the **bottom** edge of the bitmap graphic in relation to the glyph design space origin.

then it follows with:

> The originOffsetX and originOffsetY values give the placement of the bitmap graphic in relation to the standard coordinate system of the glyph design space.

So, Safari Mac has it correct, and Chrome has it wrong. In this test font, the originOffsetY is -25, which means the bottom of the bitmap should be shifted 25 units _down_ (not up!) in relation to the glyph origin.

anthrotype: I found this https://github.com/MicrosoftDocs/typography-issues/issues/191

anthrotype: The OT sbix spec also says:

> If there is a glyph contour, the glyph design space origin for the graphic is placed at the lower left corner of the glyph bounding box (xMin, yMin)

So, since we need to change nanoemoji to emit the bbox lower-left and top-right points anyway (otherwise Chrome Mac displays all blank) and hence we have to have some non-empty glyf contour that fully covers the bitmap width/height, then we could just set originOffsetX/Y == 0,  because we want the bitmap to be placed exactly "at the lower left corner of the glyph bounding box (xMin, yMin)". Hopefully everything aligns to full integers, unless upem (or ascender - descender, actually, we use that to scale the bitmap) and bitmap pixel size are not exactly divisible.

anthrotype: FreeType still hasn't been fixed, it's applying originOffsetY in the wrong direction (y-down), and it's not taking into account whether a glyf entry esists for the color glyph (in which case the glyf bbox lower left point should be used as the origin for the originOffsetX/Y):
https://gitlab.freedesktop.org/freetype/freetype/-/issues/998

This is the line where it sets the `FT_Glyph_Metrics.horiBearingY` (distance from baseline to the top of the bitmap) from sbix `originOffsetY` (note the minus sign): https://gitlab.freedesktop.org/freetype/freetype/-/blob/master/src/sfnt/ttsbit.c#L1590

`metrics->horiBearingY = (FT_Short)( -originOffsetY + metrics->height );`

Now I wonder, what if we do add the non-empty glyf entry with lower-left and top-right invisible points (to make the bitmap draw on Mac) and set originOffsetX/Y to 0 accordingly.. then FreeType would still appear incorrect because it ignores the glyf entry and just uses the baseline y=0 as the origin for originOffsetY instead of the bbox.yMin... 🤔

anthrotype: Anyway, if nanoemoji ships with any sbix support, the fonts should at least work on the Apple platform which is the canonical implementation. It may be impossible to make them work reliably everywhere because of the legacy issues above.

tests/rect_glyf_colr_1_and_picosvgz_and_sbix.ttx
rsheeter: because I passed the document through lxml to remove the inline bitmaps

rsheeter: it's harmless though a touch ugly.

rsheeter: https://stackoverflow.com/questions/25813756/lxml-kills-my-cdata-sections

src/nanoemoji/write_font.py
```diff
@@ -148,24 +160,30 @@ class ColorGenerator(NamedTuple):
...
+    # https://github.com/googlefonts/nanoemoji/issues/260 svg, colr
     # Meant to be subset if used for network delivery
     "glyf_colr_1_and_picosvgz": ColorGenerator(
         lambda *args: _colr_ufo(1, *args),
         lambda *args: _svg_ttfont(*args, picosvg=True, compressed=True),
         ".ttf",
     ),
+    # https://github.com/googlefonts/nanoemoji/issues/260 svg, colr, sbix; max compatibility
+    # Meant to be subset if used for network delivery
+    "glyf_colr_1_and_picosvgz_and_sbix": ColorGenerator(
```
rsheeter, line 181: Per f2f w/Cosimo we should be uncompressed svg and cbdt because sbix is wildly inconsistently implemented. See https://github.com/harfbuzz/harfbuzz/issues/2679.

src/nanoemoji/bitmap_tables.py
anthrotype: according to Ben, empty glyf entries may also still be an issue on Safari Mac: https://github.com/harfbuzz/harfbuzz/issues/2679#issuecomment-1021479899

rsheeter: Per https://github.com/harfbuzz/harfbuzz/issues/2679#issuecomment-1021479899 it appears we need a glyf entry even for Safari

src/nanoemoji/write_font.py
rsheeter: Oops, yes, evidence of a bad idea :D

tests/rect_glyf_colr_1_and_picosvgz_and_sbix.ttx
rsheeter: Fixed

src/nanoemoji/write_font.py
rsheeter: Ty

src/nanoemoji/bitmap_tables.py
rsheeter: Being sleepy :D round() is definitely a better choice

src/nanoemoji/bitmap_tables.py
rsheeter: Adopted the final change (originOffsetY) based on discussions such as https://github.com/harfbuzz/harfbuzz/issues/2679#issuecomment-1021419864. I suppose I shall have to abandon the notion that sbix is the bitmap format that works in the most places.

src/nanoemoji/write_font.py
rsheeter: Fixed

rsheeter: Fixed

tests/nanoemoji_test.py
```diff
@@ -210,69 +209,3 @@ def test_omit_empty_color_glyphs():
         "omit_empty_color_glyphs.ttx",
         include_tables=["GlyphOrder", "cmap", "glyf", "COLR", "SVG "],
     )
-
```
anthrotype, line 213: why did you delete this whole chunk of test file below? There are two test methods which I think are unrelated to your PR

anthrotype:
LGTM to merge after restoring the deleted tests in nanoemoji_test.py https://github.com/googlefonts/nanoemoji/pull/370#pullrequestreview-863345184

anthrotype:
also, it may be worth mentioning in the README that `cargo install resvg` is required for bitmap support

tests/nanoemoji_test.py
anthrotype: I amended your last commit to restore the deleted tests

anthrotype:
replaced `int()` with `round()` in https://github.com/googlefonts/nanoemoji/commit/5e92664b80d52082b4cff8ea543b31309414b081, and verified the `--color_format=cbdt` font matches exactly NotoColorEmoji.ttf metrics

rsheeter:
> replaced int() with round() in 5e92664, and verified the --color_format=cbdt font matches exactly NotoColorEmoji.ttf metrics

Thank you :)

> also, it may be worth mentioning in the README that cargo install resvg is required for bitmap support

Will do

PR #371, opened 2022-01-18
Make resvg available to CI

rsheeter:
To support bitmap fonts - which we seem to need in some cases - we need something that renders svgs. This makes resvg available. The `cargo install resvg` step is annoyingly slow.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,12 +33,23 @@ jobs:
           - {platform: windows-latest, python-version: 3.9}
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


anthrotype:
> The cargo install resvg step is annoyingly slow.

maybe we can cache it across builds. This article https://svartalf.info/posts/2020-04-10-github-action-for-binary-crates-installation/ contains a few tips, but I haven't tried any of them yet.

anthrotype (approved):

rsheeter:
Caching sounds promising, I'll file an issue for that.

Issue #372, opened 2022-01-19
Explore caching for crates


rsheeter:
See if https://github.com/googlefonts/nanoemoji/pull/371#issuecomment-1016732886 reference helps with build time impact of adding resvg to CI

PR #373, opened 2022-01-19
use pip to install precompiled resvg wheel

anthrotype:
Towards fixing #372

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,10 @@ jobs:
     - name: Cache resvg
       uses: actions/cache@v2
       with:
-        path: ~/.cargo/bin/resvg
+        path: |
+          ~/.cargo/bin/resvg
+          ~/.cargo/.crates.toml
+          ~/.cargo/.crates2.json
         key: ${{ runner.os }}-resvg
 
     - name: Install resvg
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,12 @@ jobs:
 
     - uses: actions-rs/cargo@v1
 
-    # this is very slow :(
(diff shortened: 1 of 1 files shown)


anthrotype:
seems to be working somehow, because cargo now compains that resvg exists already:
https://github.com/googlefonts/nanoemoji/runs/4872442013?check_suite_focus=true#step:6:10

anthrotype:
seems like it worked for linux at least: 

> Ignored package `resvg v0.20.0` is already installed

but on windows the cache never hits, and the rust build is the one taking the longest.. Maybe the cargo directory lives in a different path on windows filesystem? or maybe i'm missing something else.

yisibl:
I have developed a resvg package for use in node.js that might solve the slow build problem in CI. This package does exactly the same thing as resvg's rendering.

You can try it out, there is no command line mode yet, but I can implement it.

anthrotype:
@yisibl thanks, but I don't think using node.js is a good idea for a python project. Maybe I'll try to use https://github.com/PyO3/maturin to package up the resvg binary crate in a python wheel so I can just use pip to install.

yisibl:
@anthrotype Got it. pyo3 is great.

anthrotype:
I created a new github repo called https://github.com/anthrotype/resvg-wheels and used maturin to make resvg wheels for all platforms and uploaded them to Github Releases. So now we can do `pip install --find-links https://github.com/anthrotype/resvg-wheels/releases/tag/v0.2.0 resvg` and pip will download the right wheel from Github.

I didn't publish the wheels to PyPI just yet, I'd like to first ask the resvg authors for their consent.
But for the sake of our own test suite, I'd say this beats the slow cargo install by a great margin (from several minutes to a couple seconds) so I'd merge this for now.

PR #377, opened 2022-01-30
Diff with Chrome

### LINKED ISSUE #338: use headless browsers  to generate raster diffs of input SVGs vs compiled fonts
Currently the `--gen_svg_font_diffs` uses a [custom executable that links to skia](https://github.com/rsheeter/skia_colr) and converts a font file and string of emoji text to a raster image, which we compare against a PNG of the original input SVG as rasterised by resvg library, and image-diffed using Pillow.
Given that both [Chrome](https://developers.google.com/web/updates/2017/04/headless-chrome) and [Firefox](https://hacks.mozilla.org/2017/12/using-headless-mode-in-firefox/) support generating screenshots in headless mode, we could maybe use that to generate the rasters for both the input SVG and the corresponding color glyph from the compiled vector color font.
This would work not only for COLRv1 (via Chrome Canary), but also for OT-SVG (via FireFox).
And we would test _the actual thing_.

rsheeter:
Fixes #338

### DIFF
--- a/src/nanoemoji/nanoemoji.py
+++ b/src/nanoemoji/nanoemoji.py
@@ -41,6 +41,7 @@ import os
 from pathlib import Path
 import platform
 import re
+import shlex
 import shutil
 import subprocess
 import sys
@@ -257,8 +258,8 @@ def _chrome_command() -> str:
     cmd, validator = {
         "Linux": ("google-chrome", shutil.which),
         "Darwin": (
-            '"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"',
-            lambda s: Path(s).is_dir(),
+            "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
+            lambda s: Path(s).is_file(),
         ),
         "Windows": ("chrome", shutil.which),
     }[platform.system()]
@@ -266,7 +267,7 @@ def _chrome_command() -> str:
     if not validator(cmd):
         raise ValueError(f"{cmd} failed validation")
 
-    return cmd
+    return shlex.quote(cmd)
 
 
 def write_config_preamble(nw, font_config: FontConfig):
@@ -298,6 +299,7 @@ def write_config_preamble(nw, font_config: FontConfig):
                 f"--window-size={res},{res}",
                 "--force-device-scale-factor=1",
                 "--virtual-time-budget=1000",
+                "--enable-features=COLRV1Fonts",  # unnecessary for 98+
                 "--screenshot=$out",
                 "$in",
             )
@@ -311,7 +313,9 @@ def write_config_preamble(nw, font_config: FontConfig):
         module_rule(
             nw,
             "write_diffreport",
-            f"--lhs_dir {svg2png_dir()} --rhs_dir {font2png_dir()} --output_file $out @$out.rsp",
+            f"--lhs_dir {rel_build(svg2png_dir())} "
+            f"--rhs_dir {rel_build(font2png_dir())} "
+            "--output_file $out @$out.rsp",
             rspfile="$out.rsp",
             rspfile_content="$in",
         )
--- a/src/nanoemoji/nanoemoji.py
+++ b/src/nanoemoji/nanoemoji.py
@@ -30,6 +30,8 @@ nanoemoji $(find ~/oss/twemoji/assets/svg -name '*.svg')
 from absl import app
 from absl import flags
 from absl import logging
+from collections.abc import Iterable
+import functools
 import glob
 from nanoemoji import codepoints, config, write_font
 from nanoemoji.config import AxisPosition, FontConfig, MasterConfig
@@ -37,11 +39,12 @@ from nanoemoji.util import fs_root, rel, only, abspath
 from ninja import ninja_syntax
 import os
 from pathlib import Path
+import platform
 import re
 import shutil
 import subprocess
 import sys
-from typing import List, NamedTuple, Optional, Tuple, Set, Sequence
+from typing import List, MutableSequence, NamedTuple, Optional, Tuple, Set, Sequence
 
 
 FLAGS = flags.FLAGS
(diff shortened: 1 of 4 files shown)


src/nanoemoji/write_font2png_html.py
```diff
@@ -0,0 +1,84 @@
...
+                  // try to hit target height by adjusting size in em
+                  let currEm = parseInt(glyph.style.fontSize);
+                  let newEm = currEm * 256 / glyph.offsetHeight;
+                  glyph.style.fontSize = `${newEm}em`;
+
+
+                })();
+            </script>
+        """.replace(
+                    "FONT_LOCATION", font_file
```
anthrotype, line 75: better to use f-string interpolation instead of .replace()

anthrotype:
I had to apply the following changes to make it run on macOS.

"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" is a file, not a directory. Since it contains spaces I had to call shlex.quote before constructing the ninja rule (or should ninja take care of escaping spaces?).

Also, you want to pass relative paths to --lhs_dir and --rhs_dir flags in write_diffreport script, otherwise Chrome Mac doesn't find the PNGs. 

```diff
diff --git a/src/nanoemoji/nanoemoji.py b/src/nanoemoji/nanoemoji.py
index a849e70..59ffa1a 100644
--- a/src/nanoemoji/nanoemoji.py
+++ b/src/nanoemoji/nanoemoji.py
@@ -41,6 +41,7 @@ import os
 from pathlib import Path
 import platform
 import re
+import shlex
 import shutil
 import subprocess
 import sys
@@ -257,8 +258,8 @@ def _chrome_command() -> str:
     cmd, validator = {
         "Linux": ("google-chrome", shutil.which),
         "Darwin": (
-            '"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome"',
-            lambda s: Path(s).is_dir(),
+            "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
+            lambda s: Path(s).is_file(),
         ),
         "Windows": ("chrome", shutil.which),
     }[platform.system()]
@@ -266,7 +267,7 @@ def _chrome_command() -> str:
     if not validator(cmd):
         raise ValueError(f"{cmd} failed validation")
 
-    return cmd
+    return shlex.quote(cmd)
 
 
 def write_config_preamble(nw, font_config: FontConfig):
@@ -311,7 +312,9 @@ def write_config_preamble(nw, font_config: FontConfig):
         module_rule(
             nw,
             "write_diffreport",
-            f"--lhs_dir {svg2png_dir()} --rhs_dir {font2png_dir()} --output_file $out @$out.rsp",
+            f"--lhs_dir {rel_build(svg2png_dir())} "
+            f"--rhs_dir {rel_build(font2png_dir())} "
+            "--output_file $out @$out.rsp",
             rspfile="$out.rsp",
             rspfile_content="$in",
         )

```

src/nanoemoji/write_font2png_html.py
rsheeter: It's quite confusing to manage a f-string that is itself a container for js-interpolations like ```glyph.style.fontSize = `${newEm}em`;```, added a comment about that.

rsheeter:
> I had to apply the following changes to make it run on macOS.

Ty! Hadn't tried it on Mac yet :)

rsheeter:
Alright, I think this works well enough to review now that I have, sigh, `--enable-features=COLRV1Fonts`

anthrotype:
> --enable-features=COLRV1Fonts

I think we don't actually want to test on Chrome < 98, which contains several bugs that were fixed later. Should we simply look for Chrome Canary instead, at least until it reaches stable?

anthrotype:
If I use Canary (with the following patch), I am getting some glitchy rendering, both for the SVG->PNG (left hand side) and for the COLRv1=>PNG (right hand side). I have no idea why.

```diff
diff --git a/src/nanoemoji/nanoemoji.py b/src/nanoemoji/nanoemoji.py
index c35ca10..7e6bc08 100644
--- a/src/nanoemoji/nanoemoji.py
+++ b/src/nanoemoji/nanoemoji.py
@@ -256,12 +256,12 @@ def write_preamble(nw):
 @functools.lru_cache()
 def _chrome_command() -> str:
     cmd, validator = {
-        "Linux": ("google-chrome", shutil.which),
+        "Linux": ("google-chrome-canary", shutil.which),
         "Darwin": (
-            "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
+            "/Applications/Google Chrome Canary.app/Contents/MacOS/Google Chrome Canary",
             lambda s: Path(s).is_file(),
         ),
-        "Windows": ("chrome", shutil.which),
+        "Windows": ("chrome-canary", shutil.which),
     }[platform.system()]
 
     if not validator(cmd):
@@ -299,7 +299,6 @@ def write_config_preamble(nw, font_config: FontConfig):
                 f"--window-size={res},{res}",
                 "--force-device-scale-factor=1",
                 "--virtual-time-budget=1000",
-                "--enable-features=COLRV1Fonts",  # unnecessary for 98+
                 "--screenshot=$out",
                 "$in",
             )

```
This is in Chrome stable (Version 97.0.4692.99 (Official Build) (x86_64)):
Screenshot 2022-02-01 at 10 42 33

This in latest Canary (Version 100.0.4861.0 (Official Build) canary (x86_64))
Screenshot 2022-02-01 at 10 41 57

anthrotype:
I tried to use Chrome Beta (Version 98.0.4758.80 (Official Build) beta (x86_64)), and it works as expected, without the COLRv1 flag.
Should we make the chrome executable configurable with CLI option?

anthrotype:
or we just wait to merge this until chrome 98 stable actually rolls out to everyone, which is.. today apparently

src/nanoemoji/nanoemoji.py
```diff
@@ -222,6 +253,23 @@ def write_preamble(nw):
...
 
+@functools.lru_cache()
+def _chrome_command() -> str:
+    cmd, validator = {
+        "Linux": ("google-chrome", shutil.which),
+        "Darwin": (
+            "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
+            lambda s: Path(s).is_file(),
+        ),
+        "Windows": ("chrome", shutil.which),
```
anthrotype, line 264: have you actually tested this actually works on windows? Not a blocker from my POV if this (self-testing) feature only works on *nix

rsheeter:
Since 98 is going out nowish I'm ok just running the default for now

anthrotype (approved):

PR #379, opened 2022-02-08
Highly sophisticated notdef glyph

rsheeter:
A font without a glyph data table fails OTS. A font with no glyphs with outlines emits a 1 byte glyf table but will emerge from woff2 with a 0 byte glyf. Provide a definition of notdef to avoid getting into this situation.

Currently I've only done this when we're actually generating glyf but I suspect even when there is no glyf or cff table (say a pure SVG font) we might want to add something to increase the odds things Just Work.

### DIFF
--- a/src/nanoemoji/write_font.py
+++ b/src/nanoemoji/write_font.py
@@ -56,6 +56,7 @@ from nanoemoji import util
 import os
 from pathlib import Path
 import ufoLib2
+from ufo2ft.outlineCompiler import StubGlyph
 from picosvg.svg import SVG
 from picosvg.svg_transform import Affine2D
 from picosvg.svg_types import SVGPath
@@ -210,6 +211,10 @@ def _ufo(config: FontConfig) -> ufoLib2.Font:
     space.width = config.width
     ufo.glyphOrder = [".notdef", ".space"]
 
+    # Always the .notdef outline, even for things like a pure SVG font
+    # This decreases the odds of triggering https://github.com/khaledhosny/ots/issues/52
+    _draw_notdef(config, ufo)
+
     # use 'post' format 3.0 for TTFs, shaving a kew KBs of unneeded glyph names
     ufo.lib[ufo2ft.constants.KEEP_GLYPH_NAMES] = config.keep_glyph_names
 
@@ -361,6 +366,22 @@ def _draw_glyph_extents(
     pen.endPath()
 
     return glyph
+
+
+def _draw_notdef(config: FontConfig, ufo: ufoLib2.Font):
+    # A StubGlyph named .notdef provides a nice drawing of a notdef
+    notdefArtist = StubGlyph(
+        ".notdef",
+        config.width,
+        config.upem,
+        config.ascender,
+        config.descender,
+    )
+
+    # UFO doesn't like just sticking StubGlyph directly in place
+    glyph = ufo[".notdef"]
+    glyph.width = notdefArtist.width
+    notdefArtist.draw(glyph.getPen())
 
 
 def _glyf_ufo(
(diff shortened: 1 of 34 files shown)


anthrotype:
> Currently I've only done this when we're actually generating glyf but I suspect even when there is no glyf or cff table (say a pure SVG font) we might want to add something to increase the odds things Just Work.

We always add an empty table when color format is `picosvg` or `untouchedsvg` (glyf or CFF, depending on the file extension suffix of the --output_file flag is respectively .ttf or .otf), the table is never actually missing, it's empty but it's there.

I see you draw the notdef glyph only inside `_colr_ufo` method if format starts with `glyf`, which means it only gets drawn for glyf-flavored COLR v0 or v1 format. But my understanding was that the issue with OTS complaining about empty outline table was to do with OT-SVG font inside FireFox..
I'd expect that the notdef drawing method be moved to where we actually create our temporary UFO (inside `_ufo` method), regardless of the color/output format, be it COLR, SVG, glyf or CFF.

src/nanoemoji/write_font.py
```diff
@@ -363,6 +363,17 @@ def _draw_glyph_extents(
...
 
+def _draw_notdef(config: FontConfig, ufo: ufoLib2.Font):
+    glyph = ufo[".notdef"]
+    pen = glyph.getPen()
+    pen.moveTo((0, 0))
+    pen.lineTo((config.width, 0))
+    pen.lineTo((config.width, config.upem))
+    pen.lineTo((0, config.upem))
+    pen.endPath()
+    glyph.width = config.width
```
anthrotype, line 374: FWIW, a more canonical way to draw a .notdef glyph (i.e. stroked rectangle centered within the advance width that vertically fits ascender/descender) can be found in ufo2ft outlineCompiler:
https://github.com/googlefonts/ufo2ft/blob/main/Lib/ufo2ft/outlineCompiler.py#L1552-L1580

maybe you could borrow that one? You could even import the ufo2ft `StubGlyph` object and pass it to ufo2ft's `compile*` methods' [`notdefGlyph` parameter](https://github.com/googlefonts/ufo2ft/blob/f4f49e444bf7a1c1b02f1a98591807cb69d4015b/Lib/ufo2ft/__init__.py#L91).

anthrotype:
+1 if we allow user to draw their own .notdef glyph, we would special case it in the list of input SVGs and use that one if present, or fall back to draw one if missing

src/nanoemoji/write_font.py
rsheeter: Nice. Note to self: StubGlyph, when name=".notdef", will draw a nice notdef glyph

rsheeter: Further note to self: do this always when generating the UFO

rsheeter: Done, ty

src/nanoemoji/write_font.py
```diff
@@ -363,6 +364,22 @@ def _draw_glyph_extents(
...
+    # A StubGlyph named .notdef provides a nice drawing of a notdef
+    notdefArtist = StubGlyph(
+        ".notdef",
+        config.width,
+        config.upem,
+        config.ascender,
+        config.descender,
+    )
+
+    # UFO doesn't like just sticking StubGlyph directly in place
```
rsheeter, line 377: @anthrotype ufo seemed angry when I tried to simply assign stubglyph as the definition of .notdef so I drew it on instead.

rsheeter:
> notdef drawing method be moved to where we actually create our temporary UFO

Good catch, fixed

src/nanoemoji/write_font.py
anthrotype: All in all, I think what you did is actually better than trying to stick in an alien glyph type

anthrotype (approved):

Issue #380, opened 2022-02-09
ValueError: invalid or unsupported color string: 'currentColor'


Theelx:
When building a COLRv1 font out of any SVG using fill="currentColor", nanoemoji raises `ValueError: invalid or unsupported color string: 'currentColor'`. However, such SVGs work fine as an OpenType-SVG, as all these SVGs are used in the Pride color font at https://colorfonts.wtf/. I have included a simple reproducible example with zero.svg, from the Pride font.

```py
nanoemoji --color_format glyf_colr_1_and_picosvg ./zero.svg

I0209 18:31:05.966827 139845619483264 nanoemoji.py:611] Proceeding with 1 config(s)
I0209 18:31:05.967463 139845619483264 nanoemoji.py:569] Wrote build/Font.toml
I0209 18:31:05.967766 139845619483264 nanoemoji.py:620] Generating build.ninja
I0209 18:31:05.969572 139845619483264 nanoemoji.py:679] ninja -C /home/theel/Downloads/fonts/testcolr/build
ninja: Entering directory `/home/theel/Downloads/fonts/testcolr/build'
[1/1] /home/theel/.pyenv/versions/3.9.5/bin/python3.9 -m nanoemoji.write_font -v 0 --config_file Font.toml --fea_file Font.fea --glyphmap_file Font.glyphmap @Font.ttf.rsp
FAILED: Font.ttf 
/home/theel/.pyenv/versions/3.9.5/bin/python3.9 -m nanoemoji.write_font -v 0 --config_file Font.toml --fea_file Font.fea --glyphmap_file Font.glyphmap @Font.ttf.rsp
Traceback (most recent call last):
  File "/home/theel/Coding/nanoemoji/src/nanoemoji/write_font.py", line 767, in main
    ufo, ttfont = _generate_color_font(font_config, inputs)
  File "/home/theel/Coding/nanoemoji/src/nanoemoji/write_font.py", line 704, in _generate_color_font
    color_glyphs = tuple(
  File "/home/theel/Coding/nanoemoji/src/nanoemoji/write_font.py", line 705, in <genexpr>
    ColorGlyph.create(
  File "/home/theel/Coding/nanoemoji/src/nanoemoji/color_glyph.py", line 446, in create
    _painted_layers(
  File "/home/theel/Coding/nanoemoji/src/nanoemoji/color_glyph.py", line 316, in _painted_layers
    _paint_glyph(debug_hint, config, picosvg, context, glyph_width)
[...]

PR #381, opened 2022-02-10
Translate SVG fill='currentColor' to COLR/CPAL 0xFFFF palette index

anthrotype:
Fixes https://github.com/googlefonts/nanoemoji/issues/380


tests/currentColor.svg
```diff
@@ -0,0 +1,20 @@
...
+      c 0.120117 -112.44 91.2402 -203.561 203.681 -203.69
+      z M326.44 596.341
+      v -292.15
+      c 0.269531 -29.7402 -15.4502 -57.3301 -41.1602 -72.2803
+      s -57.4697 -14.9502 -83.1797 0
+      s -41.4199 42.54 -41.1504 72.2803
+      v 292.15
+      c 0.410156 45.4102 37.3301 82 82.7402 82
+      s 82.3398 -36.5898 82.75 -82
+      z" />
```
rsheeter, line 19: Suggest change the path to a rect or something, makes the test a bit noisy to have a long path that has nothing to do with the problem

rsheeter (approved):
I think maybe this would be better tested with asserts than ttx and then what you validate will be narrower and clearer.

PR #384, opened 2022-02-14
Fix hybrid vector+bitmap color font

### LINKED ISSUE #382: hybrid color font with both vector and bitmap tables contains invalid bitmaps
PR #370 added support for bitmap tables sbix and CBDT/CBLC. It also added support for a hybrid --color_format="glyf_colr_1_and_picosvg_and_cbdt" containing all COLRv1, SVG (vector) and CBDT (bitmap) tables.
However the bitmaps inside the CBDT table are not PNG files, but the raw bytes of the SVG files :)
The logic in nanoemoji.config.FontConfig methods `has_picosvgs` and `has_bitmaps` currently assumes that it's either or, whereas it in the case of the hybrid color font, it's actually both.

anthrotype:
Fixes #382 

Before this PR, the assumption in many parts of the codebase was that there were either vector-only or bitmap-only color formats; and for the former, the `ColorGlyph.filename` pointed to an `.svg` file, whereas for the latter, it pointed to a `.png` file. Then #370 introduced the notion of hybrid `glyf_colr_1_and_picosvg_and_cbdt` font that contains CBDT, SVG and COLR tables. However, the CBDT table was still reading the PNG bytes from the same `ColorGlyph.filename`, but the latter still pointed to an `.svg` file, hence it was invalid.

After this PR, `ColorGlyph` object gets an additional optional `bitmap` attribute containing the PNG bytes, in addition to the existing optional `svg` attribute containing a `picosvg.SVG` object. Both are optional, though at least one is required; for hybird vector+bitmap formats, both are not None.

~~The `ColorGlyph.bitmap` bytes are read in `nanoemoji.write_font._inputs` function, from the `.png` file at the same location as the input `ColorGlyph.filename`; i.e. for bitmap-only formats, this _is_ the `ColorGlyph.filename` itself; whereas for hybrid format, this is a file with the same name but different `*.png` suffix in the same directory as the (picosvg) `*.svg`. This effectively becomes an inplicit dependency, and I added them to the ninja write_font build rule so that they are taken into account when building the hybrid font.
Defining a different build subdirectory where to place the png files for the hybrid build, would have required that nanoemoji.write_font be changed to explicitly take the additional input PNG files, or alternatively some flag to translate from the *.svg to the respective *.png. All in all, my current apprach is simpler and keeps the changes to a minimum, so I went with that.~~

EDIT: After Rod's review, I reverted to adding both the svg and png filenames to the glyphmap.csv as first two columns, one of the two can be omitted (left empty). The default `write_glyphmap` script takes all the .svg and .png input files and matches them by their path stem (basename excluding the suffix) and places them on the same row, generating codepoints and glyph_name from their shared stem.
`write_font` parses the glyphmap.csv to obtain the list of svg and/or png input filenames, loads one or both depending on whether the desired config's color format `has_svgs` or `has_bitmaps` or both, and assigns them to a ColorGlyph.

### DIFF
--- a/src/nanoemoji/config.py
+++ b/src/nanoemoji/config.py
@@ -19,6 +19,7 @@ try:
 except ImportError:
     import importlib_resources as resources  # pytype: disable=import-error
 
+import itertools
 from pathlib import Path
 from picosvg.svg_transform import Affine2D
 import toml
@@ -163,23 +164,39 @@ class FontConfig(NamedTuple):
     masters: Tuple[MasterConfig, ...] = ()
     source_names: Tuple[str, ...] = ()
 
+    def _has_any(self, *color_formats) -> bool:
+        return bool(set(color_formats).intersection(self.color_format.split("_")))
+
     @property
     def output_format(self):
         return Path(self.output_file).suffix
 
     @property
-    def has_bitmaps(self):
-        return self.color_format.startswith("sbix") or self.color_format.startswith(
-            "cbdt"
+    def has_bitmaps(self) -> bool:
+        return self._has_any("sbix", "cbdt")
+
+    @property
+    def has_picosvgs(self) -> bool:
+        return self._has_any("glyf", "colr", "picosvg", "picosvgz")
+
+    @property
+    def has_untouchedsvgs(self) -> bool:
+        return self._has_any("untouchedsvg", "untouchedsvgz")
+
+    @property
+    def has_svgs(self) -> bool:
+        return self.has_picosvgs or self.has_untouchedsvgs
+
+    @property
+    def is_vf(self) -> bool:
+        return len(self.masters) > 1
+
+    @property
+    def is_ot_svg(self) -> bool:
+        return self._has_any(
+            "".join(p)
+            for p in itertools.product(("picosvg", "untouchedsvg"), ("", "z"))
         )
-
-    @property
-    def has_picosvgs(self):
-        return not (self.color_format.startswith("untouchedsvg") or self.has_bitmaps)
-
-    @property
-    def has_svgs(self):
-        return not self.has_bitmaps
 
     def validate(self):
         for attr_name in (
@@ -199,6 +216,15 @@ class FontConfig(NamedTuple):
 
         if self.clipbox_quantization is not None and self.clipbox_quantization < 1:
             raise ValueError("If set, 'clipbox_quantization' must be 1 or positive")
+
+        # sanity check
+        assert self.has_svgs or self.has_bitmaps
+
+        if self.is_vf:
+            if self.has_bitmaps:
+                raise ValueError("bitmap formats cannot have multiple masters")
+            if self.is_ot_svg:
+                raise ValueError("OT-SVG formats cannot have multiple masters")
 
         return self
 
--- a/src/nanoemoji/nanoemoji.py
+++ b/src/nanoemoji/nanoemoji.py
@@ -134,7 +134,7 @@ def _fea_file(font_config: FontConfig) -> Path:
 
 def _font_rule(font_config: FontConfig) -> str:
     suffix = "_font"
-    if _is_vf(font_config):
+    if font_config.is_vf:
         suffix = "_vfont"
     return Path(font_config.output_file).stem + suffix
 
@@ -155,14 +155,14 @@ def _ufo_config(font_config: FontConfig, master: MasterConfig) -> Path:
 
 def _glyphmap_rule(font_config: FontConfig, master: MasterConfig) -> str:
     master_part = ""
-    if _is_vf(font_config):
+    if font_config.is_vf:
         master_part = "_" + master.style_name.lower()
     return "write_" + Path(font_config.output_file).stem + master_part + "_glyphmap"
 
 
 def _glyphmap_file(font_config: FontConfig, master: MasterConfig) -> Path:
     master_part = ""
-    if _is_vf(font_config):
+    if font_config.is_vf:
         master_part = "." + master.output_ufo
     return _per_config_file(font_config, master_part + ".glyphmap")
 
@@ -189,7 +189,7 @@ def module_rule(
 
 
 def write_font_rule(nw, font_config: FontConfig, master: MasterConfig):
-    if _is_vf(font_config):
+    if font_config.is_vf:
         rule_name = _ufo_rule(font_config, master)
         config_file = _ufo_config(font_config, master)
     else:
@@ -282,7 +282,7 @@ def write_config_preamble(nw, font_config: FontConfig):
 
     for master in font_config.masters:
         write_font_rule(nw, font_config, master)
-    if _is_vf(font_config):
+    if font_config.is_vf:
         module_rule(
             nw,
             "write_variable_font",
@@ -479,15 +479,15 @@ def write_svg_font_diff_build(
 
 
 def _input_files(font_config: FontConfig, master: MasterConfig) -> List[Path]:
-    if font_config.has_bitmaps:
-        input_files = [bitmap_dest(f) for f in master.sources]
-    elif font_config.has_picosvgs:
-        input_files = [
+    input_files = []
+    if font_config.has_picosvgs:
+        input_files.extend(
             picosvg_dest(font_config.clip_to_viewbox, f) for f in master.sources
-        ]
-    else:
-
-        input_files = [abspath(f) for f in master.sources]
+        )
+    if font_config.has_untouchedsvgs:
+        input_files.extend(rel_build(f) for f in master.sources)
+    if font_config.has_bitmaps:
+        input_files.extend(bitmap_dest(f) for f in master.sources)
     return input_files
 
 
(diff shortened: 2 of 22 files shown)


src/nanoemoji/config.py
```diff
@@ -168,18 +169,39 @@ def output_format(self):
         return Path(self.output_file).suffix
 
     @property
-    def has_bitmaps(self):
-        return self.color_format.startswith("sbix") or self.color_format.startswith(
-            "cbdt"
+    def has_bitmaps(self) -> bool:
+        return bool({"sbix", "cbdt"}.intersection(self.color_format.split("_")))
```
rsheeter, line 173: Consider {sbix, cbdt} & self.color_format.split, to my understanding use of those operators is idiomatic

src/nanoemoji/config.py
```diff
@@ -168,18 +169,39 @@ def output_format(self):
...
+            {"glyf", "colr", "picosvg", "picosvgz"}.intersection(
+                self.color_format.split("_")
+            )
+        )
+
+    @property
+    def has_untouchedsvgs(self) -> bool:
+        return bool(
+            {"untouchedsvg", "untouchedsvgz"}.intersection(self.color_format.split("_"))
         )
```
rsheeter, line 187: Looking at how these methods are evolving makes me think the color format should become a class with properties rathe rthan a string. Out of scope here I suppose.

I think an internal _has would help here, then you get

```python
    def has_bitmaps(self) -> bool:
        return _has_any("sbix", "cbdt")
```

src/nanoemoji/config.py
```diff
@@ -168,18 +169,39 @@ def output_format(self):
...
+        return len(self.masters) > 1
+
+    @property
+    def is_ot_svg(self) -> bool:
+        return bool(
+            set(
+                "".join(p)
+                for p in itertools.product(("picosvg", "untouchedsvg"), ("", "z"))
+            ).intersection(self.color_format.split("_"))
+        )
```
rsheeter, line 204: isn't this the same as "svg" in color_format?

src/nanoemoji/config.py
```diff
@@ -200,6 +222,18 @@ def validate(self):
         if self.clipbox_quantization is not None and self.clipbox_quantization < 1:
             raise ValueError("If set, 'clipbox_quantization' must be 1 or positive")
 
+        # sanity check
+        assert self.has_svgs or self.has_bitmaps
+        # currently we only support picosvg-based vector+bitmap hybrid formats,
+        # no untouchedsvg+bitmap
+        assert not (self.has_bitmaps and self.has_untouchedsvgs)
```
rsheeter, line 229: I dislike this one, I don't see a benefit to this failing the day I decide I do want an untouchedsvg_sbix_pancake font?

src/nanoemoji/nanoemoji.py
```diff
@@ -479,15 +483,16 @@ def write_svg_font_diff_build(
...
         ]
-    else:
-
+    elif font_config.has_untouchedsvgs:
         input_files = [abspath(f) for f in master.sources]
+    elif font_config.has_bitmaps:
+        input_files = [bitmap_dest(f) for f in master.sources]
+    else:
+        raise AssertionError(f"Unsupported color_format: '{font_config.color_format}'")
     return input_files
```
rsheeter, line 496: I think maybe these should be just ifs, it's weird that has_picosvgs AND has_bitmaps can both be true but then we don't list the bitmaps as inputs.  You'd have to adjust how write_font receives input but ... I think that's more correct. And then you can delete the implicit deps business in write_static_font_build.

rsheeter (approved):
LGTM, just the one nit about bitmap_dest should be the SoT for bitmap destinations

src/nanoemoji/config.py
anthrotype: to use the set operators both operands need to be sets, wheras the named methods allow the second operand to be any iterable, in this case a list. That's why I went with those. The alternative would be `bool({"sbix", "cbdt"} & set(self.color_format.split("_")))` which makes a temporary set object for the sake of the intersection and throws that away. It's not more concise either.

src/nanoemoji/config.py
anthrotype: well, right now yes.. but who knows, for whatever reason, there may appear another color_format which contains the substring "svg" but doesn't produce an OT-SVG table (which is what this `is_ot_svg` property is checking). I prefer to keep it that way.

src/nanoemoji/config.py
anthrotype: I'll keep as a string for now, but I like the idea of adding the `_has_any` helper

src/nanoemoji/config.py
anthrotype: yeah.. the reason being the code expects a .png to be found next to the input .svg, which is sort of ok for picosvg format for which we create build subfolders, not very much for untouchedsvg for which we use the original input svg file. OK, you convinced to change write_font to accept explicit .png inputs as well.

src/nanoemoji/nanoemoji.py
```diff
@@ -411,11 +411,15 @@ def write_bitmap_builds(
...
+    has_picosvgs: bool,
     master: MasterConfig,
 ):
     os.makedirs(str(bitmap_dir()), exist_ok=True)
     for svg_file in master.sources:
-        dest = bitmap_dest(svg_file)
+        if has_picosvgs:
+            dest = picosvg_dest(clipped, svg_file).with_suffix(".png")
+        else:
+            dest = bitmap_dest(svg_file)
```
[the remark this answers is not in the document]
anthrotype: it was intentional, I wanted to place the bitmap for the hybrid font next to their respective picosvgs -- it does the job, but you're right it's better to have explicit inputs and keep bitmaps in their own directories.

tests/write_font_test.py
```diff
@@ -484,41 +484,55 @@ def test_picosvg(self, color_format, svgs, expected_opacity):
...
+        ("cff_colr_0", True, False),
+        ("cff_colr_1", True, False),
+        ("cff2_colr_0", True, False),
+        ("cff2_colr_1", True, False),
+        ("picosvg", True, False),
+        ("picosvgz", True, False),
+        ("untouchedsvg", True, False),
+        ("untouchedsvgz", True, False),
+        ("glyf_colr_1_and_picosvg", True, False),
+        ("glyf_colr_1_and_picosvg_and_cbdt", True, True),
```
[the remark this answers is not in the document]
anthrotype: do you mean only for this specific test parametrization? I don't think it's worth defining an enum.IntFlag just for that. And there're only two states, only one true, both true.

src/nanoemoji/nanoemoji.py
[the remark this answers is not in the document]
anthrotype: now I remember why I didn't go the explicit input *.png route. The write_font script doesn't simply take a list of svgs (it used to), but a glyphmap.csv file with svg filenames mapped to glyph names and optional codepoints. You don't want to add additional rows to this glyphmap.csv for the *.png input files, for that would risk having inconsistent mappings for the svg and png glyphs; but if you add an extra column for the .png filename, then most of the time for vector-only fonts, that column will be empty (two ugly commas next to one another separated by whitespace at best)... Not just the default write_glyphmap need changing, but also any client-provided glyphmap-generator would have to change to allow outputing the additional optional column for the .png filenames. Sure, we _can_ do it.
The alternative of having a flag that tells write_font how to turn an svg filename into the respective png, like changing the prefix to a different directory and replace the .svg suffix with .png (so that pngs can live in their separate "bitmap" build subdirectory), would not work reliably because of the dynamic disambiguation logic you added for avoiding clashes of sources with the same basename.

anthrotype:
I have something working locally which I want to clean up first, will push it tomorrow

anthrotype:
Ok, I changed it so that

1) nanoemoji.write_font now (once again) reads its glyph inputs from `argv` list of positional arguments instead of from the glyphmap file; inputs will contain .svg and/or .png files and are grouped in pairs by matching them by their path stem (the part of the basename excluding the suffix, e.g. "bar" for either "foo/bar.svg" or "foo/bar.png").
2) the glyphmap is produced from the original SVGs (before they undergo picosvg or rasterization to bitmap); this is to avoid having a .csv with differing number of columns depending on whether it has only one .svg or .png or both an .svg and a .png; what's importat for the glyphmap is the file stem, not the extension suffix, and neither the directory where the file is located;
3) write_font takes both a --glyphmap_file and a list of svg/png file arguments, and matches one with the other by keying on their stem; this ensures that, e.g., when there's both a 0061.svg and 0061.png, they get the same codepoints and glyph name mappings, and are attached to a single InputGlyph which has both an `svg` and `bitmap` attributes.

rsheeter:
> this is to avoid having a .csv with differing number of columns depending on whether it has only one .svg or .png or both an .svg and a .png

The change to blend glyphmap and argv bugs me, glyphmap was specifically supposed to be self-contained. Why not just make the CSV always contain `..., svg_file, png_file,...`? - some formats would leave one or the other blank but that's fine. If both are present and codepoints from the respective filenames don't match that's an error condition.

anthrotype:
Ok, I restored glyphmap.csv as the single source of inputs for write_font, with first two columns (optionally empty) for svg and png filenames. Also updated top description to match the current state of the PR. I think this is good to merge now.

src/nanoemoji/png.py
```diff
@@ -0,0 +1,36 @@
...
+    def __new__(cls, *args, **kwargs):
+        self = super().__new__(cls, *args, **kwargs)
+        header = self[:8]
+        if header != cls.SIGNATURE:
+            raise ValueError("Invalid PNG file: bad signature {header!r}")
+        return self
+
+    @classmethod
+    def read_from(cls, path: Union[str, os.PathLike]) -> "PNG":
+        return cls(Path(path).read_bytes())
```
rsheeter, line 36: nit: reading the whole file to get 8 bytes seems a touch generous with memory even for Python

rsheeter:
LGTM!

src/nanoemoji/png.py
anthrotype: I'm not reading it just to get 8 bytes, we are reading all bytes anyway in order to stick them in the CBDT/sbix tables, we may as well also assert that it's actually a valid PNG file.

anthrotype: Maybe you meant that we should not reading all PNGs upfront upon gathering the input glyphs, but only one by one as we build the bitmap tables glyphs, releasing them afterwards to avoid keeping them all in memory throughout the build? I did that to keep consistent with the pattern where InputGlyph as well as ColorGlyph hold both an svg filename and a preloaded SVG object, not just the filename. But we aren't doing much to the bitmaps, it's true, we can change it if your prefer.

rsheeter: Oh my bad, I read it as doing nothing other than reading the identifying bytes and missed we are hanging onto them to shove into bitmap tables. False alarm!

Issue #388, opened 2022-02-22
Support image compression for bitmap tables


rsheeter:
https://github.com/google/zopfli/blob/master/README.zopflipng can make images considerably smaller. Nice for shipping CBDT or sbix; currently our image files are **very** big. I suspect we might have excessively high quality as well. For example:

```
Listing table info for "build/NotoColorEmojiCompatTest-Regular.ttf":
    tag     checksum    length    offset
    ----  ----------  --------  --------
    CBDT  0xBB8DA37D  30225326   3482756
    CBLC  0x76F3F7E5     14828  33708084
    COLR  0x6E5EB864   1089721  33722912
    CPAL  0xC0B3EDC4     23530  34812636
    GSUB  0x8BFC763D     26672  34836168
    OS/2  0x43FEF87B        96       392
    SVG   0x1E9C01BB  23859748  34862840
    cmap  0xF9B4AED1      8214     78988
    glyf  0xBF51858B   3238177    244148
    head  0x2087B4EA        54       268
    hhea  0x08B50431        36       324
    hmtx  0x793A03E8     78498       488
    loca  0x2C18967D    156944     87204
    maxp  0x9C353564        32       360
    name  0x20463653       396   3482328
    post  0xFFB60033        32   3482724
```

Contrast with the Noto Emoji CBDT build:

```
Listing table info for "/usr/local/google/home/rsheeter/oss/noto-emoji/fonts/NotoColorEmoji.ttf":
    tag     checksum    length    offset
    ----  ----------  --------  --------
    CBDT  0x8BF55200   9784376     15024
    CBLC  0x7EB6271C     14836   9799400
    GSUB  0xE04A3DE5     45382   9814236
    OS/2  0x760467D7        96       344
    cmap  0xAABC3BDB      2865     10780
    head  0x146639B6        54       220
    hhea  0x11640DAB        36       276
    hmtx  0xD2F60000     10340       440
    maxp  0x0ECE0039        32       312
    name  0x7340A570      1344     13648
    post  0xFB270084        32     14992
    vhea  0x0E5E04CC        36   9859620
    vmtx  0x13880000      7450   9859656
```

anthrotype:
if we want to parallelize using ninja, we'd still need a CLI entry point that can be called by ninja, so all in all it's easier to use the existing zopflipng executable than wrapping it as python extension module

rsheeter:
[...]

PR #392, opened 2022-03-02
Add support for compressing PNGs using zopflipng

anthrotype:
Part of https://github.com/googlefonts/nanoemoji/issues/388

requires https://github.com/fonttools/py-zopfli/pull/17 to be merged and released first


src/nanoemoji/config.py
```diff
@@ -159,6 +162,7 @@ class FontConfig(NamedTuple):
     fea_file: str = "features.fea"
     glyphmap_generator: str = "nanoemoji.write_glyphmap"
     bitmap_resolution: int = 128
+    use_zopflipng: bool = True
```
anthrotype, line 165: zopfli is kinda slow, should we default to False here?

anthrotype:
With this PR, CBDT table for (nanoemoji-built) Noto Emoji shrinked by -35%, from 27MB to 17MB. Still some way to go to reach the size of the original CBDT font (around 9MB). I hope pngquant will help get there.

src/nanoemoji/config.py
rsheeter: True is probably the right default. Does zopflipng have a quality option? - common in compressors. If so, perhaps this is an int, default is whatever max is, and some value (-1?) turns it off

src/nanoemoji/nanoemoji.py
```diff
@@ -273,6 +273,13 @@ def write_config_preamble(nw, font_config: FontConfig):
         )
         nw.newline()
 
+        if font_config.use_zopflipng:
+            nw.rule(
+                "zopflipng",
+                f"{sys.executable} -m zopfli.png -y"
+                f'{" -v" if FLAGS.verbosity > 0 else ""} $in $out',
```
rsheeter, line 280: Doesn't seem like interpolation is helping much, and nesting a ternary in it is a bit hard to read.

`(" -v" if FLAGS.verbosity > 0 else "") + " $in $out"`

Or at cost of lines but easier to read

```
args = "$in $out"
if FLAGS.verbosity:
  args = "-v " + args
```

tests/nanoemoji_test.py
```diff
@@ -207,9 +219,15 @@ def test_build_cbdt_font():
         "compat_font/config.toml",
     ],
 )
+@pytest.mark.parametrize("use_zopflipng", [True, False])
 @pytest.mark.skipif(RESVG_PATH is None, reason="resvg not installed")
-def test_build_compat_font(config_file):
-    tmp_dir = _run((locate_test_file(config_file),))
+def test_build_compat_font(config_file, use_zopflipng):
```
rsheeter, line 224: doesn't really demonstrate zopfli'ing actually does anything? Perhaps we could run on/off and assert bitmap is smaller? - technically not guarranteed but we should be able to use a test asset where it's extremely probable

rsheeter (approved):
LGTM!

src/nanoemoji/config.py
anthrotype: there's no straightforward --quality option I'm afraid. There's the number of iterations to try (there's actually two parameters that control that, one for small files and one for larger files), where more iterations make it slower but provide slightly better compression. noto-emoji builder simply calls it with default parameters, I think it's not worth it to add options to control the iterations, the defaults already seem to be the best compromise.

src/nanoemoji/nanoemoji.py
anthrotype: Done

tests/nanoemoji_test.py
anthrotype: good point, i was lazy :)

PR #393, opened 2022-03-04
maximum_color COLR => SVG

rsheeter:
First swing at COLR to SVG. Note that I rejigged svg generation to in general slightly, in particular we now always keep glyph names until after make_svg_table which enables svg.py::_ensure_groups_grouped_in_glyph_order to rely on the names.


anthrotype:
I needed to merge #390 so you want to change the base branch from `svg2colr` to `main` now

src/nanoemoji/colr_to_svg.py
```diff
@@ -63,9 +67,8 @@ def _map_font_space_to_viewbox(
 
 
 def _svg_root(view_box: Rect) -> etree.Element:
-    svg_tree = etree.parse(
-        str(test_helper.locate_test_file("colr_to_svg_template.svg"))
-    )
+    with resources.path("nanoemoji.data", "colr_to_svg_template.svg") as template_file:
```
anthrotype, line 70: it seems a bit overkill to use importlib.resources to embed this file, it's very small and you could simply write inline as a string literal inside the code somewhere.

```python
COLR_TO_SVG_TEMPLATE = """\
<svg viewBox="TBD" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
  <defs/>
</svg>"""
```

src/nanoemoji/colr_to_svg.py
```diff
@@ -291,28 +307,46 @@ def _new_reuse_cache() -> ReuseCache:
...
 
 
-def _colr_v0_to_svgs(view_box: Rect, ttfont: ttLib.TTFont) -> Dict[str, SVG]:
+def colr_glyphs(font: ttLib.TTFont) -> Iterable[int]:
+    colr = font["COLR"]
+    if colr.version == 0:
+        for glyph_name in colr.ColorLayers:
+            yield font.getGlyphID(glyph_name)
+    else:
+        for base_glyph in font["COLR"].table.BaseGlyphList:
```
anthrotype, line 316: technically in COLRv1 there could be _both_ v0 and v1 sort of color glyphs -- even though nanoemoji doesn't produce those.

src/nanoemoji/colr_to_svg.py
rsheeter: Is importlib considered "heavy"? 

(no objection to the proposed change, just curious)

src/nanoemoji/colr_to_svg.py
rsheeter: Fair point. I think explicit failure (rather than stealth broken output) might suffice for this first swing at this.

src/nanoemoji/colr_to_svg.py
anthrotype: no, not heavy per se, it just feels to me a bit overengineering for such a short strings. E.g. for the default config.toml, it makes more sense. Anyway I don't mind if you keep it.

rsheeter: I killed it

src/nanoemoji/colr_to_svg.py
rsheeter: added asserts

anthrotype (changes requested):
with the last commit, you hit this bug/quirkiness in fonttools: 
https://github.com/fonttools/fonttools/issues/2060#issuecomment-1063932428

Some of the expected test .ttx files changed in the wrong way because of that.
I think we didn't hit this before because we were simply reshuffling the order of the base color glyphs which are all empty (contain no glyf outline) while keeping the other glyf glyphs in the same place. Whereas now you are pushing them at the end, without also updating the glyf table object's own `glyphOrder`.

The patch below should fix the problem (and then you need to revert the changes to the .ttx files)

```diff
diff --git a/src/nanoemoji/svg.py b/src/nanoemoji/svg.py
index af6820d..931839a 100644
--- a/src/nanoemoji/svg.py
+++ b/src/nanoemoji/svg.py
@@ -610,7 +610,8 @@ def _ensure_groups_grouped_in_glyph_order(
         glyph_order
     ), f"lhs only {set(ttfont.getGlyphOrder()) - set(glyph_order)} rhs only {set(glyph_order) - set(ttfont.getGlyphOrder())}"
     ttfont.setGlyphOrder(glyph_order)
-
+    assert ttfont.isLoaded("glyf")
+    ttfont["glyf"].glyphOrder = glyph_order

 def _use_href(use_el):
     ref = use_el.attrib[_XLINK_HREF_ATTR_NAME]
```

anthrotype:
When I try to run `maximum_color` and I pass a relative (non absolute) paths to a font, ninja fails because it can't find it since its current directory is the build directory. Something like this below fixes the problem:

```
--- a/src/nanoemoji/maximum_color.py
+++ b/src/nanoemoji/maximum_color.py
@@ -266,6 +266,7 @@ def _run(argv):

     if gen_ninja():
         logging.info(f"Generating {build_file.relative_to(build_dir())}")
+        input_font = rel_build(input_font)
         with open(build_file, "w") as f:
             nw = NinjaWriter(f)
             _write_preamble(nw, input_font)
```

anthrotype:
maybe we should  add a test that confirms that doing `nanoemoji --color_format glyf_colr_1_and_picosvg` gives the same (or functionally the same?) result as running `nanoemoji --color_format glyt_colr_1` or (`picosvg`) and then `maximum_color` on the previous font?

anthrotype:
what do we do with the `tools/colr2svg.py` script? Maybe we can delete it, since it's superseded by `python -m nanoemoji.generate_svgs_from_colr`?
The `colr2svg.py` script has two additional flags, one to set a custom viewbox size, and another to set a given number of decimal digits for rounding floats. Maybe you can add equivalent to `generate_svgs_from_colr` to replicate that functionality, can be handy at times. By default we'd still want it to not round floats.

src/nanoemoji/glue_together.py
```diff
@@ -34,14 +36,9 @@
...
-        if paint.Format == ot.PaintFormat.PaintGlyph:
-            _glyphs_to_copy.add(paint.Glyph)
-
-    for record in donor["COLR"].table.BaseGlyphList.BaseGlyphPaintRecord:
-        record.Paint.traverse(donor["COLR"].table, _collect_glyphs)
+    _glyphs_to_copy = {
+        p.Glyph for p in paints_of_type(donor, ot.PaintFormat.PaintGlyph)
+    }
 
     for glyph_name in _glyphs_to_copy:
```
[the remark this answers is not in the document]
rsheeter: ...am idiot, good catch :)

src/nanoemoji/colr_to_svg.py
```diff
@@ -267,20 +267,39 @@ def descend(parent: etree.Element, paint: otTables.Paint):
         raise NotImplementedError(ot_paint.Format)
 
 
+def glyph_region(ttfont: ttLib.TTFont, glyph_name: str) -> Rect:
+    return Rect(
+        0,
+        ttfont["OS/2"].sTypoDescender,
```
anthrotype, line 273: SVG goes y-down (not y-up), so the glyph_region rectangle (in font-space) that you want to map to the viewbox in SVG space also needs to go y-down, like below.
Once you apply a patch like that, you'll notice that the <g> color glyphs of the SVG created by maximum_color from a COLR input font will no longer have any `transform` attribute, because the coordinates are already in font-space and there's a 1:1 mapping between the spaces.
Also, the baseline will now correctly be at y=0 in both OT-SVG and glyf.

```diff
diff --git a/src/nanoemoji/colr_to_svg.py b/src/nanoemoji/colr_to_svg.py
index 78dbda6..a657eaa 100644
--- a/src/nanoemoji/colr_to_svg.py
+++ b/src/nanoemoji/colr_to_svg.py
@@ -57,8 +57,9 @@ ViewboxCallback = Callable[[str], Rect]  # f(glyph_name) -> Rect
 
 
 def map_font_space_to_viewbox(view_box: Rect, glyph_region: Rect) -> Affine2D:
-    ascender = glyph_region.h + glyph_region.y
-    descender = glyph_region.y
+    ascender = -glyph_region.y
+    assert ascender >= 0
+    descender = ascender - glyph_region.h
     assert descender <= 0
     width = glyph_region.w
 
@@ -270,7 +271,7 @@ def _colr_v1_paint_to_svg(
 def glyph_region(ttfont: ttLib.TTFont, glyph_name: str) -> Rect:
     return Rect(
         0,
-        ttfont["OS/2"].sTypoDescender,
+        -ttfont["OS/2"].sTypoAscender,
         ttfont["hmtx"][glyph_name][0],
         ttfont["OS/2"].sTypoAscender - ttfont["OS/2"].sTypoDescender,
     )
diff --git a/src/nanoemoji/svg.py b/src/nanoemoji/svg.py
index 0756985..23fbfb7 100644
--- a/src/nanoemoji/svg.py
+++ b/src/nanoemoji/svg.py
@@ -441,7 +441,9 @@ def _add_glyph(svg: SVG, color_glyph: ColorGlyph, reuse_cache: ReuseCache):
         raise ValueError(f"{color_glyph.svg_filename} must declare view box")
 
     # https://github.com/googlefonts/nanoemoji/issues/58: group needs transform
-    svg_g.attrib["transform"] = _svg_matrix(color_glyph.transform_for_otsvg_space())
+    transform = color_glyph.transform_for_otsvg_space()
+    if not transform.almost_equals(Affine2D.identity()):
+        svg_g.attrib["transform"] = _svg_matrix(transform)
 
     vbox_to_upem = color_glyph.transform_for_font_space()
     upem_to_vbox = vbox_to_upem.inverse()

```

src/nanoemoji/write_font.py
```diff
@@ -220,7 +220,12 @@ def _ufo(config: FontConfig) -> ufoLib2.Font:
     _draw_notdef(config, ufo)
 
     # use 'post' format 3.0 for TTFs, shaving a kew KBs of unneeded glyph names
-    ufo.lib[ufo2ft.constants.KEEP_GLYPH_NAMES] = config.keep_glyph_names
+    # if we're building svgs keep glyph names to simplify operation on the resulting binary
+    keep = config.keep_glyph_names
+    if config.has_svgs and config.has_picosvgs:
```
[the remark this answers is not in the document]
rsheeter: the idea isto end up with has_any(picosvg, picosvgz). has_picosvgs includes glyf and colr but isn't meant to activate this.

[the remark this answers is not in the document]
rsheeter: I think it's correct but I agree it *looks* wrong

tests/omit_empty_color_glyphs.ttx
```diff
@@ -49,7 +49,9 @@
 
     <TTGlyph name="B"/><!-- contains no outline data -->
 
-    <TTGlyph name="B.0" xMin="518" yMin="-10" xMax="998" yMax="470">
+    <TTGlyph name="B.0"/><!-- contains no outline data -->
```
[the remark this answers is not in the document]
rsheeter: bleh, good catch

src/nanoemoji/colr_to_svg.py
rsheeter: y-down. Must remember y-down!

src/nanoemoji/maximum_color.py
```diff
@@ -266,6 +266,7 @@ def _run(argv):
 
     if gen_ninja():
         logging.info(f"Generating {build_file.relative_to(build_dir())}")
+        input_font = input_font.resolve()  # we need a non-relative path
```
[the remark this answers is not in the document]
rsheeter: Relative to build for files outside build is pretty ugly, instead of say /my/dir/file.ttf I get a build file with ../../..[etc]/my/dir/file.ttf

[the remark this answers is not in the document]
rsheeter: (I think using absolute for files that aren't under build is probably good in general)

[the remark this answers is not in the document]
rsheeter: Resolving but feel free to reopen if you feel strong relative is right; I'm not too attached to absolute

rsheeter:
@anthrotype I see warnings for "GSUB/GPOS Coverage is not sorted by glyph ids." ... is there by any chance a convenient method that fixes that for me?

anthrotype:
If the GSUB has been fully decompiled before reordering glyphs, then upon recompile the Coverage will be ordered using the new glyph order, so it's weird you see that. I'll take a look next week (I'm off today)

anthrotype:
if you'd like to run OTS on the CI to check fonts work for web, you can use the python wrapper that we maintain at https://github.com/googlefonts/ots-python, available from PyPI as `pip install opentype-sanitizer`. The entry point (equivalent to the C++ `ots-sanitize` executable) is `python -m ots --help`

src/nanoemoji/maximum_color.py
```diff
@@ -116,13 +151,62 @@ def _write_preamble(nw: NinjaWriter, input_font: Path):
...
+
+def _generate_svg_from_colr(nw: NinjaWriter, input_font: Path, font: ttLib.TTFont):
+    # generate svgs
+    svg_files = [
+        rel_build(svg_generate_dir() / f"{gid:05d}.svg") for gid in colr_glyphs(font)
+    ]
+    nw.build(svg_files, "generate_svgs_from_colr", input_font)
+    nw.newline()
+
+    # picosvg them
```
anthrotype, line 176: i think we could get rid of this step because the SVGs extracted by colr_to_svg.py are already (or should already be) "picosvg"

src/nanoemoji/glue_together.py
```diff
@@ -58,13 +56,45 @@ def _collect_glyphs(paint):
...
+    # To keep things simple, let's build a new glyph order that keeps all the svg font gids stable
+    target_glyph_order = list(target.getGlyphOrder())
+    svg_glyphs = {gn for _, gn in _svg_glyphs(donor)}
+    non_svg_target_glyphs = [gn for gn in target_glyph_order if gn not in svg_glyphs]
+    new_glyph_order = []
+
+    for svg_gid, svg_glyph_name in _svg_glyphs(donor):
+        # we want gid to remain stable so copy non-svg glyphs until that will be true
+        while len(new_glyph_order) < svg_gid:
+            new_glyph_order.append(non_svg_target_glyphs.pop(0))
```
anthrotype, line 76: I think this while loop is never actually triggered because all the svg glyphs are placed in a contiguous block of gids at the end (see `nanoemoji.svg._ensure_groups_grouped_in_glyph_order`). Maybe an assert bould be just fine.
This glue_together module is for internal use only and the "donor" font is one that nanoemoji itself has built.

src/nanoemoji/maximum_color.py
anthrotype: Nevermind actually, we do emit g elements which are not pico. Let's keep for now

anthrotype (approved):

anthrotype:
i marked "approved" but it's not ready yet until the issue with glyph reordering in #399 is fixed first

rsheeter:
I think this is sufficiently far along to be worth merging. I expect follow-on PRs but I don't anticipate tossing the core approach at this point.

src/nanoemoji/glue_together.py
```diff
@@ -58,13 +58,45 @@ def _collect_glyphs(paint):
...
+    target = load_fully(Path(FLAGS.target_font))
+    donor = load_fully(Path(FLAGS.donor_font))
 
+    # TODO lookup, guess fn name, etc
     if FLAGS.color_table == "COLR":
         _copy_colr(target, donor)
+    elif FLAGS.color_table == "SVG":
+        _copy_svg(target, donor)
     else:
+        # TODO: SVG support
```
anthrotype, line 97: i believe this TODO no longer applies, isn't the point of the current PR to add COLR=>SVG support?
also the following comment lines seem redundant to me

src/nanoemoji/maximum_color.py
```diff
@@ -116,13 +151,62 @@ def _write_preamble(nw: NinjaWriter, input_font: Path):
...
+
+    # generate a new font with SVG glyphs that use the same names as the original
+    nw.build(
+        "svg_from_colr.ttf",
+        "write_svg_font_from_generated_svgs",
+        [_COLR2SVG_GLYPHMAP, _COLR2SVG_CONFIG],
+    )
+    nw.newline()
+
+    # stick our shiny new COLR table onto the input font and declare victory
```
anthrotype, line 200: we actually stick "our shiny new _SVG_ table onto the input font", not COLR

tests/rect_glyf_colr_1_and_picosvg_and_cbdt.ttx
```diff
@@ -72,15 +72,15 @@
   <CBDT>
     <header version="3.0"/>
     <strikedata index="0">
-      <cbdt_bitmap_format_17 name="e000">
+      <cbdt_bitmap_format_17 name="e000.0">
```
anthrotype, line 75: hm this has the wrong glyph name, it should be "e000" (the base color glyph) not "e000.0" (one of COLR layers)..

src/nanoemoji/glue_together.py
anthrotype: oh I see what's going on here. Even if we know all the svg color glyphs are in a contiguous block in the donor's glyphOrder (which nanoemoji itself has built), we don't control the target's glyphOrder which can be anything; and at the same time, we don't want to have to remap the `id="glyphX"` attributes in the OT-SVG documents. So when glueing together the two fonts make sure the glyph ids of for the SVG glyphs to stay the same as in the donor font and reshuffle the target's glyphs around accordingly.

PR #394, opened 2022-03-08
Add support for quantizing PNGs with pngquant

anthrotype:
The pngquant command is now available as 'pngquant-cli' on PyPI so we can install it with pip like the rest of the dependencies: https://pypi.org/project/pngquant-cli

I still had to wrap the pngquant command with a script that handles two special error codes 98 (output larger than input) and 99 (conversion quality below minimum) by copying the input file to the output file. The [noto-emoji Makefile](https://github.com/googlefonts/noto-emoji/blob/9a5261d871451f9b5183c93483cbd68ed916b1e9/Makefile#L186) has a similar rule. If we don't add that, then a bunch of noto-emojis would fail to quantize (e.g. the standalone skin tone modifiers e.g. 🏽), as ninja sees the non-zero exit code and stops early.

I tried running this on my local machine and I now get a CBDT noto-emoji font which is around 10MB large (only 220KB larger than the original, not sure exactly where the extra bytes come from but I think it's a good start).

I'm going to add some tests that check that everything runs as expected

### DIFF
--- a/src/nanoemoji/nanoemoji.py
+++ b/src/nanoemoji/nanoemoji.py
@@ -472,6 +472,8 @@ def _input_files(font_config: FontConfig, master: MasterConfig) -> List[Path]:
             dest_func = zopflipng_dest
         elif font_config.use_pngquant:
             dest_func = pngquant_dest
+        # zopflipng always happens after pngquant, so when both are true
+        # the final desired file is zopflipng_dest
         input_files.extend(dest_func(f) for f in master.sources)
     return input_files
 
--- a/src/nanoemoji/nanoemoji.py
+++ b/src/nanoemoji/nanoemoji.py
@@ -227,10 +227,10 @@ def write_config_preamble(nw, font_config: FontConfig):
 
         if font_config.use_pngquant:
             # always overwrite using --force
-            pngquant_args = f"-f {font_config.pngquant_flags} -o $out $in"
+            pngquant_args = f"-f {font_config.pngquant_flags}"
             if FLAGS.verbosity:
                 pngquant_args = "-v " + pngquant_args
-            nw.rule("pngquant", f"pngquant {pngquant_args}")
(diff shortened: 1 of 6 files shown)


src/nanoemoji/nanoemoji.py
```diff
@@ -374,16 +391,22 @@ def write_bitmap_builds(
...
+    builds: Set[Path],
+    nw: NinjaWriter,
+    master: MasterConfig,
+    rule_name: str,
+    dest_dir: Path,
+    infile_fn: Callable[[Path], Path],
+    outfile_fn: Callable[[Path], Path],
 ):
-    os.makedirs(str(zopflipng_dir()), exist_ok=True)
+    os.makedirs(str(dest_dir), exist_ok=True)
```
rsheeter, line 403: Optional, there's a whole set of required dirs created early in the build, could create this there

src/nanoemoji/nanoemoji.py
```diff
@@ -599,18 +626,51 @@ def _run(argv):
...
-                    if font_config.has_bitmaps:
-                        write_bitmap_builds(
-                            bitmap_builds, nw, font_config.clip_to_viewbox, master
-                        )
+                if font_config.has_bitmaps:
+                    assert not font_config.is_vf
+                    write_bitmap_builds(
+                        bitmap_builds,
+                        nw,
+                        font_config.clip_to_viewbox,  # currently unused
```
rsheeter, line 634: Why not remove?

rsheeter (approved):

rsheeter:
Give or take 200KB sounds good enough to me

src/nanoemoji/nanoemoji.py
anthrotype: but these are only required if any of the font configs has bitmaps and uses either zopflipng or pngquant, etc. better to create on demand i think

PR #399, opened 2022-03-16
Implement a more complete reorder glyphs

rsheeter:
Per f2f we might need to sort a bunch of things in || with coverage. Playing with mechanism to facilitate this.

See https://github.com/fonttools/fonttools/issues/2060 for context.

### DIFF
--- a/src/nanoemoji/reorder_glyphs.py
+++ b/src/nanoemoji/reorder_glyphs.py
@@ -1,0 +1,71 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Helps with glyph reordering.
+
+See https://docs.google.com/document/d/14b8bivUBdWSkeGLIVcqVLBVWNbXXFawguq52D2olvEA/edit for context."""
+
+
+from fontTools import ttLib
+from fontTools.ttLib.tables import otTables as ot
+from nanoemoji.util import bfs_base_table, require_fully_loaded, SubTablePath
+from typing import List
+
+
+_COVERAGE_ATTR = "Coverage"  # tables that have one coverage use this name
+
+
+# (Type, Optional Format) => List of (attr name of Coverage, optional attr name of list sorted by that coverage)
+# See Cosimo's doc for context
+_REORDER_NEEDED = {
+    (ot.SinglePos, 1): [(_COVERAGE_ATTR, None)],
+    (ot.SinglePos, 2): [(_COVERAGE_ATTR, "Value")],
+    (ot.PairPos, 1): [(_COVERAGE_ATTR, "PairSet")],
+    (ot.PairPos, 2): [(_COVERAGE_ATTR, None)],
+}
+
+
+def _access_path(path: SubTablePath):
+    path_parts = []
+    for entry in path:
+        if not entry.name:
+            continue
+        path_part = entry.name
+        if entry.index is not None:
+            path_part += f"[{entry.index}]"
+        path_parts.append(path_part)
+    return ".".join(path_parts)
+
+
+def reorder_glyphs(font: ttLib.TTFont, new_glyph_order: List[str]):
+    # Changing the order of glyphs in a TTFont requires that all tables that use
+    # glyph indexes have been fully decompiled (loaded with lazy=False).
+    # Cf. https://github.com/fonttools/fonttools/issues/2060
+
+    require_fully_loaded(font)
+
+    font.setGlyphOrder(new_glyph_order)
+    # glyf table is special and needs its own glyphOrder...
+    font["glyf"].glyphOrder = new_glyph_order
+
+    # TODO also GSUB, GDEF, MATH ... see Cosimo's Doc. For now just GPOS for testing purposes.
+    coverage_containers = {"GPOS"}
+    for tag in coverage_containers:
+        if tag in font.keys():
+            for path in bfs_base_table(font[tag].table):
+                print(_access_path(path))
+                value = path[-1].value
+                reorder_key = (type(value), getattr(value, "Format", None))
+                if reorder_key in _REORDER_NEEDED:
+                    print("OMG need to reorder", reorder_key)
--- a/src/nanoemoji/reorder_glyphs.py
+++ b/src/nanoemoji/reorder_glyphs.py
@@ -21,10 +21,38 @@ from fontTools import ttLib
 from fontTools.ttLib.tables import otBase
 from fontTools.ttLib.tables import otTables as ot
 from nanoemoji.util import bfs_base_table, require_fully_loaded, SubTablePath
-from typing import List, NamedTuple, Optional
+from typing import Any, Callable, List, NamedTuple, Optional
 
 
 _COVERAGE_ATTR = "Coverage"  # tables that have one coverage use this name
+
+
+def _sort_by_gid(
+    get_glyph_id: Callable[[str], int],
+    glyphs: List[str],
+    parallel_list: Optional[List[Any]],
+):
+    if parallel_list:
+        reordered = sorted(
+            ((g, e) for g, e in zip(glyphs, parallel_list)),
+            key=lambda t: get_glyph_id(t[0]),
+        )
+        sorted_glyphs, sorted_parallel_list = map(list, zip(*reordered))
+        parallel_list[:] = sorted_parallel_list
+    else:
+        sorted_glyphs = sorted(glyphs, key=get_glyph_id)
+
+    glyphs[:] = sorted_glyphs
+
+
+def _get_dotted_attr(value: Any, dotted_attr: str) -> Any:
+    attr_names = dotted_attr.split(".")
+    assert attr_names
+
+    while attr_names:
+        attr_name = attr_names.pop(0)
+        value = getattr(value, attr_name)
+    return value
 
 
 class ReorderCoverage(NamedTuple):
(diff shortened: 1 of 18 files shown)


src/nanoemoji/util.py
```diff
@@ -85,30 +87,120 @@ def file_printer(filename):
...
         if tag == "cmap":
             _ = [st.cmap for st in table.tables]
-        ttfont[tag] = table
 
+    _assert_all_keys_loaded(font)
+
+    # post has a cool pattern of raising errors if it has no glyphOrder attribute
+    # font isn't all that useful w/o glyph order so lets confirm we're ok on that front
+    # TODO contemplate whether this is simply a FontTools bug
+    font.getGlyphOrder()
```
anthrotype, line 132: Why this?

src/nanoemoji/util.py
```diff
@@ -85,30 +87,120 @@ def file_printer(filename):
...
+
+    # make sure reverse glyph order rebuilds too
+    font.getReverseGlyphMap(rebuild=True)
+
+    # TEMPORARY HOPEFULLY
+    # The glyphs in Coverage observably do not update properly, following logs:
+    # otTables.py:581] GSUB/GPOS Coverage is not sorted by glyph ids.
+    # Despite Coverage attempting self-repair resulting font will fail OTS.
+
+    # TODO also GSUB?
```
anthrotype, line 199: Yes, all of GPOS, GSUB, GDEF and MATH can contain Coverages in various places

src/nanoemoji/util.py
```diff
@@ -85,30 +87,120 @@ def file_printer(filename):
...
+    font2 = ttLib.TTFont(tmp, lazy=lazy)
+
+
+def load_fully(font: Union[Path, ttLib.TTFont]) -> ttLib.TTFont:
+    if isinstance(font, Path):
+        font = ttLib.TTFont(str(font), lazy=False)
+    else:
+        # A TTFont might be opened lazily and some tables only partially decompiled.
+        # If so, reload it
+        if font.lazy:
```
anthrotype, line 111: `lazy` has three states, True, False and None (which is the default and means something in between). I know, very confusing; it should be documented better.
```suggestion
        if font.lazy is not False:
```

src/nanoemoji/util.py
```diff
@@ -85,30 +87,120 @@ def file_printer(filename):
...
+    while frontier:
+        # path is (value, attr_name) tuples. attr_name is attr of parent to get value
+        path = frontier.pop(0)
+        current = path[-1].value
+        visited.add(id(current))
 
-def reorder_glyphs(ttfont: ttLib.TTFont, glyph_order: List[str]):
+        callback(path)
+
+        converters = getattr(current, "converters", [])
```
anthrotype, line 156: BaseTable has a `getConverters` public method, you should use that instead of reaching directly to the converters yourself

src/nanoemoji/util.py
```diff
@@ -85,30 +87,120 @@ def file_printer(filename):
...
+                print("  ", "^has no name")
+                raw_value = None
+
+            if not isinstance(raw_value, list):
+                raw_value = [raw_value]
+
+            for value in raw_value:
+                if (
+                    value is not None
+                    and hasattr(value, "converters")
```
anthrotype, line 174: you should check that `isinstance(value, BaseTable)`, having an attribute named "converters" may be too broad. This traverse method is meant for trees of BaseTables

src/nanoemoji/util.py
```diff
@@ -85,30 +87,120 @@ def file_printer(filename):
...
+    _assert_all_keys_loaded(font)
+
+    font.setGlyphOrder(glyph_order)
     # glyf table is special and needs its own glyphOrder...
-    assert ttfont.isLoaded("glyf")
-    ttfont["glyf"].glyphOrder = glyph_order
+    font["glyf"].glyphOrder = glyph_order
+
+    # make sure reverse glyph order rebuilds too
+    font.getReverseGlyphMap(rebuild=True)
```
anthrotype, line 192: this is not needed. as soon as you do `TTFont.setGlyphOrder`, the `self._reverseGlyphOrderDict` cache is deleted so next time one gets the reversed glyph map it's gonna be rebuilt

src/nanoemoji/util.py
```diff
@@ -85,30 +87,120 @@ def file_printer(filename):
...
+
+    # TEMPORARY HOPEFULLY
```
anthrotype, line 194: well not temporary, it's just the way otData-tables work in fonttools; there are parallel lists (various Coverage and linked arrays) that need to be kept in the same order; changing the glyph order requires one to change the order of those tables as well, as we are doing below. It's more likely that fonttools grows an additional API to reorder glyphs which transparently does all this work for us, than Coverage and linked lists updating automatically on compile as the glyphOrder changes. So I think that warning is not gonna go away. The attempt at self-repairing is some obscure hack that was put in place for some SIL fonts
https://github.com/fonttools/fonttools/blob/71a5bf1b512e51044fe2af31d86f2e73bd3a3eaa/Lib/fontTools/ttLib/tables/otTables.py#L527-L529
I'm not even sure it ever worked tbh

src/nanoemoji/util.py
```diff
@@ -85,30 +87,120 @@ def file_printer(filename):
...
+            if not isinstance(raw_value, list):
+                raw_value = [raw_value]
+
+            for value in raw_value:
+                if (
+                    value is not None
+                    and hasattr(value, "converters")
+                    and id(value) not in visited
+                ):
+                    frontier.append(path + (OtDataPathEntry(value, attr_name),))
```
anthrotype, line 177: knowing the position (index) of value in an array of something (for "repeat" converters) in addition to the path leading to it, could be useful as well. e.g. knowing that a given `MarkBasePos` object is at `gpos.LookupList.Lookup[32].SubTable[2]`

src/nanoemoji/util.py
anthrotype: the reason you are getting garbage for the format-switching subtables is that accessing the inner `converters` attribute gives you a dict (keyed by format) not a list in those cases. You want to use `getConverters()` instead which returns the correct list of converters all the time for both normal and format-switching subtables

anthrotype:
calling `getattr(current, "converters", [])` is wrong because it doesn't take into account FormatSwiching tables (Coverage itself, all the lookup subtables, etc.) which contain a mapping of converters keyed by the format. Even using the public `getConverters()` method (which is supposed to resolve the format and return a list of converters) may still raise AttributeError for some format-swiching tables that hide their original "Format". Again, Coverage is an example, but also SingleSubst, and other simple substitution subtables. These have custom subclasses of BaseTable in the otTables.py module that provide a more high-level API than regular otData-driven classes. The decision to delete the Format attributes from these was a somewhat controversial design decision by Behdad (which caused some failures downstream), the rationale being fonttools decompile either formats to the same high-level data and takes care of compiling to the optimal format transparently. These sort of tables seem to always appear at the end of a the class tree as leaves, and never contain other tables so your traversal could simply skip trying to traverse their children when getCoverters() can't give you an answer because that table doesn't remember what Format it used to be..

```diff
diff --git a/src/nanoemoji/util.py b/src/nanoemoji/util.py
index a6d95a5..a96f176 100644
--- a/src/nanoemoji/util.py
+++ b/src/nanoemoji/util.py
@@ -17,6 +17,7 @@
 import contextlib
 from fontTools.ttLib.tables import otTables
 from fontTools.ttLib.tables import otConverters
+from fontTools.ttLib.tables.otBase import BaseTable
 from fontTools import ttLib
 from functools import partial
 from io import BytesIO
@@ -153,7 +154,17 @@ def traverse_ot_data(root, callback: Callable[[OtDataPath], None]) -> Iterable[A

         callback(path)

-        converters = getattr(current, "converters", [])
+        try:
+            converters = current.getConverters()
+        except AttributeError as e:
+            # FormatSwitchingBaseTable.getConverters() may fail for some tables that
+            # hide their real "Format" (e.g. Coverage, SingleSubst, etc. - search
+            # fontTools' otTables.py for "del self.Format"). We can skip traversing
+            # these further for these are 'leaves' containing simple, non-otData-driven
+            # data (e.g. a list of glyphs or a mapping of glyphs to substitue)
+            if "Format" == e.args[0]:
+                continue
+
         for converter in converters:
             is_simple = isinstance(converter, otConverters.SimpleValue)
             print("  ", getattr(converter, "name", "?nameless?"), converter, is_simple)
@@ -171,7 +182,7 @@ def traverse_ot_data(root, callback: Callable[[OtDataPath], None]) -> Iterable[A
             for value in raw_value:
                 if (
                     value is not None
-                    and hasattr(value, "converters")
+                    and isinstance(value, BaseTable)
                     and id(value) not in visited
                 ):
                     frontier.append(path + (OtDataPathEntry(value, attr_name),))
```

src/nanoemoji/util.py
```diff
@@ -85,30 +87,120 @@ def file_printer(filename):
...
+def _reload(font: ttLib.TTFont, lazy: bool = True):
+    # Stream font to memory and load it back again
     tmp = BytesIO()
-    ttfont.save(tmp)
+    font.save(tmp)
     tmp.seek(0)
-    ttfont2 = ttLib.TTFont(tmp, lazy=False)
-    for tag in ttfont2.keys():
-        table = ttfont2[tag]
+    font2 = ttLib.TTFont(tmp, lazy=lazy)
```
anthrotype, line 102: you are not returning the `font2` here

src/nanoemoji/util.py
anthrotype: note that my old `ensure_ttfont_fully_decompiled` was called unconditionally before, not just when lazy. The rationale was that the TTFont we receive from ufo2ft might be in an in-between state that is resolved only upon serializing the font (to disk or memory) because something else happens or is computed upon compile, e.g. the post table format may be set to 3 to drop the glyph names but the rest of the loaded tables still use the old names and it's only upon compile that the names are actually stripped away; so we'd compile the font to a temporary stream and reload off that.

src/nanoemoji/util.py
rsheeter: The comment aimed to explain but I guess it doesn't. It's possible to create a font where getGlyphOrder will fail after a reorder attempt and it's really annoying when that happens later in some random spot. I prefer to confirm it won't kerplode right here.

src/nanoemoji/util.py
rsheeter: QQ, should we just delete the current self-repair stuff? - it's very confusing that it looks like it fixes itself but ... doesn't.

rsheeter: (and given your clarification of things sorted in parallel I'm not sure it knows enough to ever do it correctly)

src/nanoemoji/util.py
anthrotype: can you tell me exactly how to reproduce this? It's certainly a bug that we should fix upstream

anthrotype: this comment

> post has a cool pattern of raising errors if it has no glyphOrder attribute

is misleading. One is not supposed to ever access the post.glyphOrder directly, only via TTFont.getGlyphOrder()

rsheeter: Ah that's my mistake in phrasing. So what ends up happening is you call TTFont::getGlyphOrder and it crashes _in_ post. It's not caused by a direct call into post.

anthrotype: then that's even more of a bug!

rsheeter: Now if only I'd been smart enough to record how to achieve that state...

src/nanoemoji/util.py
rsheeter: Oops, that's a stale comment from before you clarified the need to sort things in ||

src/nanoemoji/util.py
rsheeter: No rush. I believe your doc explains sufficiently to build at least the outlines of a correct solution we can then iterate on here and hopefully in time promote into FT.

src/nanoemoji/util.py
anthrotype: in https://github.com/fonttools/fonttools/pull/2551/files#diff-fb09acb555593d2fddbfc5bc3bf295f7727e027cdf97d023ab095a02ef2c9436R860, I also added an `iterSubTables` method to iterate over all the BaseTable-children of a given BaseTable (if any) -- I needed that to call ensureDecompiled recursively to fully decompile GPOS/GSUB from a lazy font. You could use that method to extend your traversal frontier.

anthrotype: changed the iterSubTables to return a namedtuple(name, value, index) so you better can fit into your traversal

https://github.com/fonttools/fonttools/pull/2551/commits/ab8fc321a76a960322a39f92cf9895066bb52458

I guess I'm gonna merge that PR before it gets too big

src/nanoemoji/util.py
```diff
@@ -85,30 +87,120 @@ def file_printer(filename):
...
+        # A TTFont might be opened lazily and some tables only partially decompiled.
+        # If so, reload it
+        if font.lazy:
+            font = _reload(font, lazy=False)
+
+    # one would think lazy=False meant tags would report isLoaded true
+    # one would be wrong, following fails:
+    # _assert_all_keys_loaded(font)
+
+    # OK, for those of you that are STILL lazy just bloody load
```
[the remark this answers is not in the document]
rsheeter: I'm going to update here assuming that lands

src/nanoemoji/util.py
rsheeter: I think maybe this doesn't matter as I've migrated to the new toys you added to fontTools. If I find a new repro I'll file a bug.

src/nanoemoji/util.py
rsheeter: Migrated to your shiny new FontTools iterSubTables

src/nanoemoji/util.py
rsheeter: Now using the new machinery, it's quite helpful.

src/nanoemoji/util.py
rsheeter: Good catch

src/nanoemoji/util.py
rsheeter: Fixed via migration to iterSubTables

src/nanoemoji/util.py
rsheeter: So close.

src/nanoemoji/reorder_glyphs.py
```diff
@@ -0,0 +1,71 @@
...
+def reorder_glyphs(font: ttLib.TTFont, new_glyph_order: List[str]):
+    # Changing the order of glyphs in a TTFont requires that all tables that use
+    # glyph indexes have been fully decompiled (loaded with lazy=False).
+    # Cf. https://github.com/fonttools/fonttools/issues/2060
+
+    require_fully_loaded(font)
+
+    font.setGlyphOrder(new_glyph_order)
+    # glyf table is special and needs its own glyphOrder...
+    font["glyf"].glyphOrder = new_glyph_order
```
anthrotype, line 60: this won't be needed as https://github.com/fonttools/fonttools/pull/2544 is merged

src/nanoemoji/reorder_glyphs.py
```diff
@@ -0,0 +1,71 @@
...
+    (ot.SinglePos, 2): [(_COVERAGE_ATTR, "Value")],
+    (ot.PairPos, 1): [(_COVERAGE_ATTR, "PairSet")],
+    (ot.PairPos, 2): [(_COVERAGE_ATTR, None)],
+}
+
+
+def _access_path(path: SubTablePath):
+    path_parts = []
+    for entry in path:
+        if not entry.name:
```
anthrotype, line 42: it's impossible for entry to not have a name (str), object attributes must have a non-empty str name

src/nanoemoji/util.py
anthrotype: it's not just "A TTFont might be opened lazily and some tables only partially decompiled." As I explained in my previous comment, it's also possible that the TTFont object is in an in-between state preceding the compile step during which some stuff may get updated or flushed away, think about post v3 glyph names dropped or glyph bounding boxes recalcuated - none of these are not related to the lazy flag.
I think you should either run `_reload` unconditionally irrespective of font.lazy (if we want to support that in-between case) or we don't run at all and simply rely on the new ensureDecompiled() to make sure that any tables that have not been loaded yet will be. Running _reload only for font.lazy = True is either unnecessary or incomplete.

anthrotype: i think you can remove the _reload altogether and if we do find issues we address them

anthrotype: the "A TTFont might be opened lazily and some tables only partially decompiled" is already taken care of by the subsequent `ensureDecompiled()`

src/nanoemoji/util.py
```diff
@@ -85,30 +87,66 @@ def file_printer(filename):
...
+    while frontier:
+        # path is (value, attr_name) tuples. attr_name is attr of parent to get value
+        path = frontier.pop(0)
+        current = path[-1].value
+
+        yield path
+
+        new_frontier = []
+        for subtable_entry in current.iterSubTables():
+            if isinstance(subtable_entry.value, otBase.BaseTable):
```
anthrotype, line 149: iterSubTables is designed/guaranteed to yield instances of BaseTable you don't need this check

src/nanoemoji/reorder_glyphs.py
anthrotype: oh you constructed a SubTableEntry("", root) below, that's what this if is catching

src/nanoemoji/util.py
```diff
@@ -85,30 +87,66 @@ def file_printer(filename):
...
+
+def _traverse_ot_data(
+    root: otBase.BaseTable, add_to_frontier_fn: AddToFrontierFn
+) -> Iterable[SubTablePath]:
+    # no visited because general otData is forward-offset only and thus cannot cycle
+
+    frontier: List[SubTablePath] = [(otBase.BaseTable.SubTableEntry("", root),)]
+    while frontier:
+        # path is (value, attr_name) tuples. attr_name is attr of parent to get value
+        path = frontier.pop(0)
```
anthrotype, line 142: How about using `deque.popleft` which is O(1) instead of list.pop(0) which is O(n) in CPython (lists are array-based) https://docs.python.org/3/library/collections.html#deque-objects

src/nanoemoji/util.py
```diff
@@ -85,30 +87,66 @@ def file_printer(filename):
...
+        current = path[-1].value
+
+        yield path
+
+        new_frontier = []
+        for subtable_entry in current.iterSubTables():
+            if isinstance(subtable_entry.value, otBase.BaseTable):
+                new_frontier.append(path + (subtable_entry,))
+
+        frontier = add_to_frontier_fn(frontier, new_frontier)
```
anthrotype, line 152: updating the deque in-place with extend/extendleft might be more efficient than creating a whole new list from two lists

src/nanoemoji/write_font.py
```diff
@@ -784,6 +784,9 @@ def _generate_color_font(config: FontConfig, inputs: Iterable[InputGlyph]):
     ttfont = _make_ttfont(config, ufo, color_glyphs)
 
     if ttfont is not None:
+        # apply_ttfont may wish to do things like reorder tables that require full load
+        ttfont = util.load_fully(ttfont)
```
anthrotype, line 788: wouldn't it be better to only do that if we are going to actually reorder glyphs? it's a bit expensive operation (especially if we do keep the reload step)

anthrotype:
FYI released https://github.com/fonttools/fonttools/releases/tag/4.31.0

src/nanoemoji/util.py
rsheeter: Will do

src/nanoemoji/util.py
rsheeter: done

src/nanoemoji/util.py
rsheeter: Fixed

src/nanoemoji/reorder_glyphs.py
rsheeter: Updated to pass font["tag"].table as the first entries accessor

src/nanoemoji/write_font.py
rsheeter: Currently we don't know if the next step reorders. I suppose we could push down the reorder and have apply_ttfont return a font that replaces the one passed in.

src/nanoemoji/reorder_glyphs.py
```diff
@@ -0,0 +1,205 @@
...
+    return ".".join(path_parts)
+
+
+def reorder_glyphs(font: ttLib.TTFont, new_glyph_order: List[str]):
+    # Changing the order of glyphs in a TTFont requires that all tables that use
+    # glyph indexes have been fully.
+    # Cf. https://github.com/fonttools/fonttools/issues/2060
+    require_fully_loaded(font)
+
+    font.setGlyphOrder(new_glyph_order)
```
garretrieger, line 195: Looking at the range request implementation of glyph reordering: https://github.com/litherum/StreamableFonts/blob/master/Optimizer/ReorderFont.py#L98

That scripts del's a couple of cached copies of the original glyph ordering, do you need to do that as well?

rsheeter: I believe not as setGlyphOrder handles both of those issues (https://github.com/fonttools/fonttools/blob/main/Lib/fontTools/ttLib/ttFont.py#L453). I believe it has long handled reverse and the handling of glyf is new.

rsheeter:
@garretrieger suggested adding a test of actually reordering a font for at least one case with parallel reordering, will do that then merge

garretrieger (approved):

PR #400, opened 2022-04-03
Add CBDT to maximum_color

### LINKED ISSUE #401: Support proportional bitmaps
Discovered during #400 that if you specify `--color_format cbdt --width=0` the results are not terribly satisfying. hmtx reports only 0, etc. Prior to #400 a crash would occur.

CBDT support was initially written to only support square inputs but there is no real reason this needs to persist.

rsheeter:
See https://github.com/googlefonts/nanoemoji/tree/maximum_bitmap#adding-color-tables-to-existing-fonts for outline of what/why.

Remove formats with multiple color tables from --color_formats, user should use maximum_color instead.

Add support for variable width bitmaps. Fixes #401.


src/nanoemoji/glue_together.py
```diff
@@ -34,7 +37,15 @@
 flags.DEFINE_string("target_font", None, "Font assets are copied into.")
 flags.DEFINE_string("donor_font", None, "Font from which assets are copied.")
 flags.DEFINE_string("color_table", None, "The color table to copy.")
-flags.DEFINE_string("output_file", None, "Font assets are copied into.")
+flags.mark_flag_as_required("target_font")
```
anthrotype, line 40: why did you move these outside the `if __name__ == "__main__":` branch? I think that will make it raise as soon as one tries to import glue_together module. I think as long as these flags are only used inside main() function, it makes sense to only require them `if __name__ == "__main__"`

src/nanoemoji/glue_together.py
```diff
@@ -84,28 +95,105 @@ def _copy_svg(target: ttLib.TTFont, donor: ttLib.TTFont):
...
+    new_order = sorted(donor_order, key=target.getGlyphID)
+
+    # now we know the desired order, reshard into runs
+    # TODO duplicative of make_cbdt_table
+    strike_template = copy.deepcopy(donor["CBLC"].strikes[0])
+    clbc_index_template = copy.deepcopy(strike_template.indexSubTables[0])
+    strike_template.indexSubTables = []
+
+    cblc = donor["CBLC"]
+    cblc.strikes = []
```
anthrotype, line 137: why is this modifying the donor instead of copying?

src/nanoemoji/glue_together.py
```diff
@@ -84,28 +95,105 @@ def _copy_svg(target: ttLib.TTFont, donor: ttLib.TTFont):
...
+    return {
+        glyph_name: CbdtGlyphInfo(data[glyph_name], sizes[glyph_name])
+        for glyph_name in data
+    }
+
+
+def _copy_cbdt(target: ttLib.TTFont, donor: ttLib.TTFont):
+    cbdt_glyph_info = _cbdt_data_and_sizes(donor)
+
+    # reorder the bitmap table to match the targets glyph order
```
anthrotype, line 120: i wonder if we should upstream to fonttools' CBDT/CBLC table modules all this logic to automatically sort the bitmap glyphs upon compile given the current (possibly updated) glyph order

src/nanoemoji/nanoemoji.py
```diff
@@ -209,7 +209,7 @@ def _chrome_command() -> str:
 
 def write_config_preamble(nw, font_config: FontConfig):
     if font_config.has_bitmaps:
-        # TODO bitmap support for non-square svgs? - baby steps
+        # conveniently if w != h this already scales w
```
anthrotype, line 212: wait, what does that mean exactly?

When w != h, what we want from resvg is that the resulting bitmap always has its height set to the desired `font_config.bitmap_resolution` (e.g. 128) whereas the bitmap width can be variable and scaled accordingly.
Remember the relative "size" of our glyhps is determined by the viewBox height which is scaled to fit the sum of ascender - descender.

I just tried to experiment passing resvg some SVGs with viewbox w != h and passing -w and -h to the same value (like you are doing here), and it looks like resvg is just scale-to-fitting to the maximum of either width or height.
We don't want that actually. If a proportional glyph is wider than tall, we want the bitmap height set to the bitmap resolution and the bitmap width set wider relative to its aspect ratio; we don't want the bitmap width set to the resolution and the height scaled down.

Actually I tried to compile a proportional cbdt font from a couple of SVGs, where one has a viewbox width that is larger than its height (e.g. one has viewBox="0 0 10 10" and the other "0 0 20 10"), and I got this error (seems like you only tested with "narrow" proportional glyphs) 

```
  File "/Users/clupo/Github/nanoemoji/src/nanoemoji/bitmap_tables.py", line 213, in _make_cbdt_strike
    bitmap_pixel_height = only({c.bitmap.size[1] for c in color_glyphs})
  File "/Users/clupo/Github/nanoemoji/src/nanoemoji/util.py", line 34, in only
    assert next(it, None) is None
AssertionError
ninja: build stopped: subcommand failed.
```

anthrotype: I believe what we want is fix only the height to resvg -h, and let resvg adjust the bitmap width proportionately, i.e. `f"resvg -h {res} $in $out"`.

anthrotype: argh.. if I fix the height and let the width free to expand, I reach a point where the latter can't fit any longer in the ridiculously tight uint8 limits of CBDT SmallGlyphMetrics!

https://docs.microsoft.com/en-us/typography/opentype/spec/cbdt#smallglyphmetrics

Seriously, does it mean one can't have a PNG in CBDT table if height or width exceed 255?
Wow..

anthrotype:
> Remove formats with multiple color tables from --color_formats, user should use maximum_color instead.

what if one only wants COLRv1 and OT-SVG but not CBDT, or only COLRv1 and CBDT but not OT-SVG? Maybe maximum_color should have an option to select the desired list of formats (defaulting to all possible ones given input).

src/nanoemoji/bitmap_tables.py
```diff
@@ -49,54 +53,85 @@
...
 
-def _width_in_pixels(config: FontConfig) -> int:
-    return round(
-        config.bitmap_resolution * config.width / (config.ascender - config.descender)
-    )
+def _pixels_to_funits(config: FontConfig, bitmap_pixel_height: int) -> Tuple[int, int]:
+    # the bitmap is vertically scaled to fill the space desc to asc
+    # this gives us a ratio between pixels and upem
+    funits = config.ascender - config.descender
+    return (bitmap_pixel_height, funits)
```
anthrotype, line 118: the function is named `_pixels_to_funits`, so i'd expect it to actually return the ratio `pixels/funits` (flaot) instead of passing them through as a tuple. It's not doing much besides subtracting ascender-descender

src/nanoemoji/bitmap_tables.py
```diff
@@ -49,54 +53,85 @@
...
-    advances = {
-        ttfont["hmtx"].metrics[ttfont.getGlyphName(c.glyph_id)][0] for c in color_glyphs
-    }
-    assert len(advances) == 1, "Proportional bitmaps not supported yet"
-    return next(iter(advances))
+    assert width_funits > 0
+    return round(width_funits * pixels / funits)
+
+
+# https://github.com/googlefonts/noto-emoji/blob/9a5261d871451f9b5183c93483cbd68ed916b1e9/third_party/color_emoji/emoji_builder.py#L53
```
anthrotype, line 131: is the link above still relevant? the code below probably diverged enough to bear little resemblance with it.

src/nanoemoji/bitmap_tables.py
```diff
@@ -118,17 +153,15 @@ def _cbdt_bitmapdata_offsets(
...
     bitmap_data = CbdtBitmapFormat17(b"", None)
     bitmap_data.metrics = SmallGlyphMetrics()
-    bitmap_data.metrics.height = config.bitmap_resolution
-    bitmap_data.metrics.width = config.bitmap_resolution
-    # center within advance
+    bitmap_data.metrics.width, bitmap_data.metrics.height = image_data.size
     bitmap_data.metrics.BearingX = metrics.x_offset
     bitmap_data.metrics.BearingY = metrics.y_offset
-    bitmap_data.metrics.Advance = _width_in_pixels(config)
+    bitmap_data.metrics.Advance = bitmap_data.metrics.width
```
anthrotype, line 164: this is wrong, we know the advance width in pixel, it's `_width_in_pixels`; the `bitmap_data.metrics.width` is "the number of columns of data", i.e. the pixel width of the actual bitmap, the advance can be larger (and in fact it is larger in our default config which mimics noto-emoji's).

```suggestion
    bitmap_data.metrics.Advance = _width_in_pixels(config, image_data)
```

anthrotype: the first frame is colrv1, the second is cbdt (from `main` branch, with correct advance) and the third frame is cbdt from this PR branch (with advance too narrow):

https://user-images.githubusercontent.com/6939968/161588016-51a410c4-3f92-4dbd-b3b7-8134e567a3c8.mov

README.md
```diff
@@ -26,6 +26,27 @@ Requires Python 3.7 or greater.
...
+
+Given at least one vector color table (COLR or SVG) the other vector color table and bitmap table(s)
+can be generated:
+
+```shell
+# Adds COLR to a font with SVG and vice versa
+maxmium_color my_colr_font.ttf
+```
+
+The intended result is a font that will Just Work in any modern browser:
```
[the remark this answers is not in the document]
rsheeter: Oh good point. @drott is there any chance we could fix that?

src/nanoemoji/bitmap_tables.py
rsheeter: Maybe I should make a type. I'm under the impression that by delaying the computation and always doing multiple (enbiggen) then divide we minimize float damage.

src/nanoemoji/bitmap_tables.py
rsheeter: ...no, not it is not. I think perhaps just a link at the top to emoji_builder.py might suffice.

src/nanoemoji/bitmap_tables.py
rsheeter: Bleh good catch, ty.

src/nanoemoji/glue_together.py
rsheeter: I copy the first strike from donor to use as a template, then rebuild a new set of strikes (see clbc.strikes.append a few lines down) based on the reordered glyphs. I'll update the comment.

rsheeter: ...and indeed the copy template is pointless since I then free all the references. Removing.

src/nanoemoji/nanoemoji.py
rsheeter: > I believe what we want is fix only the height to resvg -h, and let resvg adjust the bitmap width proportionately, i.e. f"resvg -h {res} $in $out".

Quite right ty, I definitely only tested narrow so far. Will update.

> does it mean one can't have a PNG in CBDT table if height or width exceed 255

I believe so, yes. If we agreed on which way y goes we'd use sbix but ... no such luck.

src/nanoemoji/glue_together.py
rsheeter: Tbh I don't remember. I'll move them back until a reason comes to mind lol.

README.md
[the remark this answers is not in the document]
rsheeter: Added a note to readme wrt chrome preferences

src/nanoemoji/bitmap_tables.py
```diff
@@ -25,22 +25,26 @@
...
 from nanoemoji.color_glyph import ColorGlyph
+from nanoemoji.png import PNG
+from nanoemoji.util import only
 from typing import (
     List,
     NamedTuple,
     Sequence,
     Tuple,
 )
+import statistics
```
anthrotype, line 42: this seems unused?

src/nanoemoji/nanoemoji.py
```diff
@@ -210,8 +210,6 @@ def _chrome_command() -> str:
 def write_config_preamble(nw, font_config: FontConfig):
     if font_config.has_bitmaps:
         # set height only, let width scale proportionally
-        # for CBDT width can fairly easily overrun tight pixel size limits
-        # but there isn't much we can do about it
```
anthrotype: you could leave this comment in -- even if there isn't much we can do about it, at least it attests that we are aware of the limitation.
I suppose the only feasible solution in these cases would be to reduce the --bitmap_resolution until it works..

src/nanoemoji/glue_together.py
anthrotype: you removed the copies so the transplant kills the donor and organs can't be reused for a different target, I suppose it's fine since the method is private and only used inside main. might be worth mentioning that in a comment maybe.

src/nanoemoji/maximum_color.py
```diff
@@ -197,16 +248,21 @@ def _generate_svg_from_colr(nw: NinjaWriter, input_font: Path, font: ttLib.TTFon
     )
     nw.newline()
 
-    # stick our shiny new COLR table onto the input font and declare victory
+    # stick our shiny new COLR table onto the input font
```
anthrotype, line 251:
```suggestion
    # stick our shiny new SVG table onto the input font
```

src/nanoemoji/write_glyphmap_for_glyph_svgs.py
```diff
@@ -35,17 +36,30 @@ def main(argv):
...
 
     with util.file_printer(FLAGS.output_file) as print:
-        for input_file in input_files:
+        while input_files:
+            svg_file = input_files.pop()
+            assert svg_file.suffix in {".png", ".svg"}, f"What is {svg_file}"
+            bitmap_file = None
+            if svg_file.suffix == ".png":
+                bitmap_file = svg_file
+                svg_file = input_files.pop()
```
anthrotype, line 51: you're relying on the fact that .svg and .png are listed one after the other, and that .png comes before .svg in the reversed list of arguments (because "p" is before "s" in alphabetical order). How about you only pass either the .svg or the .png instead of both, since we are only interested in the file.stem anyway.

anthrotype: oh actually seems like you are using both the svg_file and bitmap_file, not just their stem..

anthrotype: nevermind, the files are sorted like that because we sorted them immediately before, all good

src/nanoemoji/glue_together.py
```diff
@@ -84,19 +91,101 @@ def _copy_svg(target: ttLib.TTFont, donor: ttLib.TTFont):
...
+        glyph_run = new_order[:end]
+        new_order = new_order[end:]
+        max_gid = target.getGlyphID(glyph_run[-1])
+
+        strike = copy.deepcopy(strike_template)
+        strike.bitmapSizeTable.min_gid = min_gid
+        strike.bitmapSizeTable.max_gid = max_gid
+
+        max_width = max(cbdt_glyph_info[gn].data.metrics.width for gn in glyph_run)
+        strike.bitmapSizeTable.hori.widthMax = max_width
```
anthrotype, line 158: in bitmap_tables.py you define widthMax as 
```python
line_metrics.widthMax = max(d.metrics.Advance for d in data.values())
```

you should do the same here I think, no?
metrics.width is the width of the bitmap, not the "advance" width.. So confusing, I know.
Also below for vertical

src/nanoemoji/maximum_color.py
```diff
@@ -238,14 +294,62 @@ def _generate_colr_from_svg(nw: NinjaWriter, input_font: Path, font: ttLib.TTFon
...
+    )
+    nw.newline()
+
+    # make a config
+    nw.build(_CBDT_CONFIG, "write_cbdt_config", input_font)
+    nw.newline()
+
+    # generate a new font with CBDT glyphs that use the same names as the original
+    # hidden in the rule definition is that the both vector table font is an input
+    # so list it as implicit
```
anthrotype, line 336: is that true that the ".both_vector_tables.ttf" font is hidden in the "write_cbdt_font" rule definition? I don't see that..
The "write_cbdt_font" simply calls write_font.py with the cbdt glyphmap and cbdt config to produce a cbdt.ttf. No mention of the ".both_vector_tables.ttf" font.

however it _is_ included in the `copy_cbdt` rule definition (which calls glue_together.py), maybe that's what you meant?

src/nanoemoji/maximum_color.py
```diff
@@ -272,9 +376,15 @@ def _run(argv):
...
+                color_font, picosvg_files = _generate_svg_from_colr(
+                    nw, input_font, font
+                )
             else:
-                _generate_colr_from_svg(nw, input_font, font)
+                color_font, picosvg_files = _generate_colr_from_svg(
+                    nw, input_font, font
+                )
+
+            _generate_cbdt(nw, input_font, font, color_font, picosvg_files)
```
anthrotype, line 387: Why to make a CBDT table from it do we need to run picosvg on the svg extracts? Can't we use the svg as they are. Resvg is definitely capable to rasterise them. I don't see the reason to do that. It just adds an extra layer of potential bugs. The two `_generate_*` methods above should just return the extracted SVGs (or converted from COLR) without (or before) running picosvg on them.

anthrotype:
> Remove formats with multiple color tables from --color_formats, user should use maximum_color instead.

I liked that one could make a hybrid font with multiple tables directly from the same set of input SVGs. Now one would have to first make either a COLRv1 or OT-SVG font, and the run maximum_color on it to add the rest of the tables (and may also find oneself with additional tables that were not requested, e.g. if one is only interested in testing vector color formats and is not interested in the bitmaps).
Not only this process is slower and doesn't take advantage of parallel ninja builds, but also it may introduce slight differences due to, e.g. converting cubic2quads or rounding to UPEM integer grid (for COLR+glyf), or picosvg doing all sorts of relatively lossy transformations to simplify the SVGs.
All this to say it was a nice feature to have and may be nice to keep.

src/nanoemoji/glue_together.py
rsheeter: Rename donor to victim? :D

In this context I think it's fine, donors sole purpose is to become somewhat one with target ;)

rsheeter:
> I liked that one could make a hybrid font with multiple tables directly from the same set of input SVGs

I think if we want that back we should add a flag to nanoemoji to directly list the color table(s) desired and then it can share code with maximum color. The method of provisioning flags for every combination of formats just didn't scale, many valid combinations were left out.

src/nanoemoji/maximum_color.py
rsheeter: Because I'd like the generated tables to be consistent. For an ot-svg input I think odds are nanoemoji would do a better job (since most existing ones seem to do svg per glyph with no sharing) so arguably we should rebuild the svg table in that situation.

Resolving but plz reopen if that argument is absurd.

src/nanoemoji/glue_together.py
rsheeter: I would love to have a functional reorder method in fonttools. My thought was to let it bake a little here first?

src/nanoemoji/bitmap_tables.py
rsheeter: oops, I think I had a statistics.mean at some point and then removed it. Fixed.

src/nanoemoji/nanoemoji.py
rsheeter: I'm going to add an explicit check to make_cbdt_table so you get an actual error

src/nanoemoji/glue_together.py
rsheeter: Good catch, fixed

src/nanoemoji/maximum_color.py
rsheeter: Hm ... I think it might be off by one indeed.

src/nanoemoji/colr_to_svg.py
```diff
@@ -296,10 +296,13 @@ def glyph_region(ttfont: ttLib.TTFont, glyph_name: str) -> Rect:
     """The area occupied by the glyph, NOT factoring in that Y flips.
 
     map_font_space_to_viewbox handles font +y goes up => svg +y goes down."""
+    width = ttfont["hmtx"][glyph_name][0]
+    if width == 0:
+        width = ttfont["glyf"][glyph_name].xMax
```
anthrotype, line 307: hm I'm not sure about using the xMax as replacement for advance width when this is 0... Are you doing this because of the clipping of bitmaps that overflow the viewbox #402? It feels like you are special-casing the case when a glyph bounds exceed horizontally.. Ideally some more global fix would be desirable.

src/nanoemoji/colr_to_svg.py
```diff
@@ -95,7 +96,12 @@ def _draw_svg_path(
 
 
 def _color(ttfont: ttLib.TTFont, palette_index, alpha=1.0) -> colors.Color:
-    cpal_color = ttfont["CPAL"].palettes[0][palette_index]
+    palette = ttfont["CPAL"].palettes[0]
+    if palette_index == _FOREGROUND_COLOR_INDEX:
+        return colors.Color.fromstring("black")  # as good a guess as any
```
anthrotype, line 101: why black? Should we not use the special "currentColor" here?
https://github.com/googlefonts/nanoemoji/pull/381

src/nanoemoji/colr_to_svg.py
rsheeter: Maybe it's a bug in the input but it seemed like it happened for glyf components

src/nanoemoji/colr_to_svg.py
rsheeter: We should!

src/nanoemoji/maximum_color.py
```diff
@@ -191,118 +184,136 @@ def _write_preamble(nw: NinjaWriter, input_font: Path):
...
-) -> Tuple[Path, List[Path]]:
-    # generate svgs
-    svg_files = [
-        rel_build(svg_generate_dir() / f"{gid:05d}.svg") for gid in colr_glyphs(font)
-    ]
-    nw.build(svg_files, "generate_svgs_from_colr", input_font)
+
+def _write_font(nw: NinjaWriter, output_file: Path, inputs: WriteFontInputs):
+    nw.build(
+        output_file, "write_font", implicit=list(inputs), variables=inputs._asdict()
```
anthrotype, line 214: ninja rule variables, nice!

src/nanoemoji/maximum_color.py
```diff
@@ -358,33 +367,38 @@ def _run(argv):
...
-                    nw, input_font, font
-                )
+                wip_file, picosvg_files = _generate_colr_from_svg(nw, wip_file, font)
+
+            if FLAGS.bitmaps:
+                wip_file = _generate_cbdt(nw, input_file, font, wip_file, picosvg_files)
 
-            _generate_cbdt(nw, input_font, font, color_font, picosvg_files)
+            # until now we held glyph names stable, nuke 'em
+            _strip_glyph_names(nw, wip_file, final_output)
```
anthrotype, line 402: glyph names may be useful for debugging so ideally maximum_color should respect the --keep_glyph_names flag (flase by default) like the main nanoemoji compiler?

anthrotype (approved):

src/nanoemoji/colr_to_svg.py
rsheeter: Tracking resolution of this in #404

src/nanoemoji/maximum_color.py
rsheeter: they work so well!

src/nanoemoji/maximum_color.py
rsheeter: Oops, I forgot we had a flag for that. Good catch!

src/nanoemoji/colr_to_svg.py
rsheeter: Tracking resolution of this in https://github.com/googlefonts/nanoemoji/issues/405

README.md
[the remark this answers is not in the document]
drott: See https://bugs.chromium.org/p/skia/issues/detail?id=12945 - discussed this with Ben and it would need changes in FreeType to block it from looking at `CBDT/CBLC`.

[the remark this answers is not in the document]
anthrotype: thanks for the update

PR #403, opened 2022-04-08
Support colr glyph => svg

### LINKED ISSUE #398: support PaintColrGlyph when doing COLR => SVG
I tried running `maximum_color` script (from PR #393) on ArefRuqaaInk-Regular.ttf from https://github.com/aliftype/aref-ruqaa (which contains a COLRv1 table) and it fails with:

```
  File "/Users/clupo/Github/nanoemoji/src/nanoemoji/colr_to_svg.py", line 269, in _colr_v1_paint_to_svg
    raise NotImplementedError(ot_paint.Format)
NotImplementedError: 11
```

This is because nanoemoji/color_to_svg.py doesn't support `PaintColrGlyph` paint yet (PaintFormat 11 is "PaintColrGlyph"), because nanoemoji doesn't emit that Paint format itself.

However, the ArefRuqaaInk font is generated using fontmake, which in turn uses glyphsLib to convert Glyphs.app' color glyphs attributes to fonttools compatible COLR Paint structures (stored in the UFO lib).

See here how they use PaintColrGlyph to encode colored composite glyphs:
https://github.com/googlefonts/glyphsLib/blob/609b1765096b016e7382b1155c4034888a0de878/Lib/glyphsLib/builder/color_layers.py#L155

It should be straightforward to add support for this, perhaps even in the same #393 PR.

rsheeter:
Fixes #398 by shamelessly copying @anthrotype's proposed solution

### DIFF
--- a/tests/colr_to_svg_test.py
+++ b/tests/colr_to_svg_test.py
@@ -1,0 +1,113 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from fontTools.colorLib.builder import buildCPAL, LayerListBuilder
+from fontTools import ttLib
+from fontTools.ttLib.tables import _g_l_y_f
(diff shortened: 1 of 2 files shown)


tests/colr_to_svg_test.py
```diff
@@ -0,0 +1,145 @@
...
+              <defs/>
+              <path d="M10,10 L90,10 L90,90 L10,90 Z"/>
+            </svg>
+            """,
+        ),
+        # Paint colr glyph
+        (
+            "paint_colr_glyph",
+            {
+                "paint_colr_glyph": (ot.PaintFormat.PaintColrGlyph, "color_box"),
```
anthrotype, line 72: an even more interesting test case would be a PaintColrGlyph inside a PaintTransform -- which, incidentally, is what glyhpsLib produces for composite color glyphs found in .glyphs3 sources (cf. ArefRuqaInk)

anthrotype (approved):
LGTM!

PR #407, opened 2022-04-19
WIP Precompute reusable parts for each picosvg

rsheeter:
This facilitates selecting the "best" shape to keep from the set that normalize the same. Intuition is that may mean the biggest one :)


src/nanoemoji/parts.py
```diff
@@ -0,0 +1,138 @@
...
+class ShapeSet:
+    normalized: Shape
+    shapes: Set[Shape] = dataclasses.field(default_factory=set)
+
+
+@dataclasses.dataclass
+class ReuseableParts:
+    version: Tuple[int, int, int] = (1, 0, 0)
+    reuse_tolerance: float = dataclasses.field(default_factory=_default_tolerence)
+    shape_sets: MutableMapping[Shape, ShapeSet] = dataclasses.field(
```
anthrotype, line 64: isn't it redundant that the normalized shape is both the key of the shape_sets dictionary and an attribute of the `ShapeSet` class? is the latter needed, or used in absence of the key that is associated with the sets of shapes?
E.g. instead of `for shape_set in source.shape_sets.values():` (and then use `shape_set.normalized`) you could equally do:

`for normalized, shape_set in source.shape_sets.items():`

even the `ShapeSet` class feels redundant to me, you could reduce everything to a `MutableMapping[str, Set[str]]` just like your json representation basically

src/nanoemoji/parts.py
```diff
@@ -0,0 +1,138 @@
...
+        self.shape_sets[norm].shapes.add(shape)
+
+    def _add(self, shape: Shape):
+        norm = shape
+        if self.reuse_tolerance != -1:
+            norm = Shape(normalize(SVGPath(d=shape.d), self.reuse_tolerance).d)
+        self._add_norm_path(norm, shape)
+
+    def add(self, source: PathSource):
+        if isinstance(source, Path):
```
anthrotype, line 80: accepting a filesystem `Path` as well feels a bit over-overloading to me, `parts.add(ReusableParts.load("path/to/parts.json"))` should be ok

src/nanoemoji/parts.py
```diff
@@ -0,0 +1,138 @@
...
+            source = ReuseableParts.load(source)
+
+        if isinstance(source, ReuseableParts):
```
anthrotype, line 83: I wonder if it would it be too clever to also override the `__or__`, `__ior__` magic methods to be able to do `parts = p1 | p2` or `parts |= other_parts`

then i'd leave the `add` method for single SVGShapes

anthrotype (approved):

src/nanoemoji/parts.py
```diff
@@ -0,0 +1,138 @@
...
+    try:
+        val = next(it)
+        return isinstance(val, desired_type)
+    except StopIteration:
+        return True
+
+
+@dataclasses.dataclass(frozen=True)
+class Shape:
+    d: str  # an SVG style path, e.g. the d attribute of <svg:path/>
```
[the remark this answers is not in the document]
rsheeter: SVGPath isn't hashable. I didn't realize there was a NewType mechanism in python, will switch to that.

src/nanoemoji/parts.py
rsheeter: I will see if I can use a NewType again. I'm not a fan of `Map<str, Set<str>>`, those types get very opaque very quickly.

rsheeter: NewType works well. Brace yourself for my burst of NewType enthusisam leading to NewType's everywhere.

src/nanoemoji/parts.py
rsheeter: Updating so.

src/nanoemoji/parts.py
rsheeter: I waffled on whether that would be pythonic and ultimately felt it was a little on the obtuse side. WDYT?

(I might clean up other feedback and merge, we can always make this change later if we like it)

src/nanoemoji/parts.py
```diff
@@ -0,0 +1,138 @@
...
+        }
+        return json.dumps(json_dict, indent=2)
+
+    @classmethod
+    def fromstring(cls, string) -> "ReuseableParts":
+        first = string.strip()[0]
+        parts = cls()
+        if first == "<":
+            svg = SVG.fromstring(string).topicosvg()
+            for path in svg.xpath("//svg:path"):
```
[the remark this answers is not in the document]
rsheeter: Forgot we had a wrapper around that, updated.

PR #410, opened 2022-04-21
Migrate to parameterized rules using ninja variables

### LINKED ISSUE #397: use ninja rule variable expansion
currently we define similar rules multiple times with different names for different flags, we can use rule variables and have build target assign values to them:  see https://ninja-build.org/manual.html#_variable_expansion

rsheeter:
Fixes #397 and resolves bugs where it looks like we could have emitted the same rule name with different parameters.

### DIFF
--- a/src/nanoemoji/nanoemoji.py
+++ b/src/nanoemoji/nanoemoji.py
@@ -52,7 +52,18 @@ import shlex
 import shutil
 import subprocess
 import sys
-from typing import Callable, List, NamedTuple, Optional, Tuple, Set, Sequence
+from typing import (
+    Any,
+    Callable,
+    List,
+    Mapping,
+    MutableMapping,
+    NamedTuple,
+    Optional,
+    Tuple,
+    Set,
+    Sequence,
+)
 
 
 FLAGS = flags.FLAGS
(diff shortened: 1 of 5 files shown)


src/nanoemoji/nanoemoji.py
```diff
@@ -131,42 +122,42 @@ def _glyphmap_file(font_config: FontConfig, master: MasterConfig) -> Path:
...
+        "Linux": ("google-chrome", shutil.which),
+        "Darwin": (
+            "/Applications/Google Chrome.app/Contents/MacOS/Google Chrome",
+            lambda s: Path(s).is_file(),
+        ),
+        "Windows": ("chrome", shutil.which),
+    }[platform.system()]
+
+    if not validator(cmd):
+        raise ValueError(f"{cmd} failed validation")
```
anthrotype, line 156: perhaps "Google Chrome not found" is clearer than "failed validation"

tests/nanoemoji_test.py
```diff
@@ -38,14 +40,26 @@
 )
 
 
-@pytest.fixture(scope="module", autouse=True)
+# https://docs.pytest.org/en/latest/example/simple.html#making-test-result-information-available-in-fixtures
+@pytest.hookimpl(tryfirst=True, hookwrapper=True)
```
anthrotype, line 44: That's fancy!
Alternatively, we could also simply switch to use the default `tmp_path` pytest fixture. It keeps temporary directories from the previous 3 test runs and removes the older ones on a rolling basis.
I refrained from using that to reduce the amount of diffs.
https://docs.pytest.org/en/6.2.x/tmpdir.html

tests/nanoemoji_test.py
```diff
@@ -38,14 +40,26 @@
...
+def pytest_runtest_makereport(item, call):
+    # execute all other hooks to obtain the report object
+    outcome = yield
+    result = outcome.get_result()
+    setattr(item, result.when + "_result", result)
+
+
+@pytest.fixture()
```
anthrotype, line 52: If you want to keep this, I believe you need to continue using `autouse=True` like before. We want the fixture to be automatically used (not only when a test function references it explicitly as a parameter). I think right now it is not being triggered.

anthrotype (approved):

tests/nanoemoji_test.py
rsheeter: Erk good catch

tests/nanoemoji_test.py
rsheeter: I tried leaving temp files for an entire test run and it's quite a lot so for now I prefer the "silent when successful" model.

tests/nanoemoji_test.py
rsheeter: Ha, you are right. And that revealed my hook wasn't quite working as intended. I think it's now happy.

Issue #402, opened 2022-04-04
Support unclipped viewbox for bitmap formats


anthrotype:
Currently, it's impossible to produce a bitmap color font (CBDT or sbix) containing color glyphs that "kern" over the next or preceding one, or that stretch beyond the ascender/descender lines. 
E.g. take this test file [0053.svg](https://user-images.githubusercontent.com/6939968/161565935-6ea8f8d0-c922-472e-9ca1-9416c4acb273.svg) and compile to CBDT. Notice how the --noclip_to_viewbox is essentially ignored:

```
$ nanoemoji --noclip_to_viewbox --color_format cbdt 0053.svg 
```
<img width="257" alt="Screen Shot 2022-04-04 at 15 27 02" src="https://user-images.githubusercontent.com/6939968/161565663-d0691069-2717-46fc-ba2f-1ba12db2815c.png">

The expected result should look like the COLRv1 version (OT-SVG also works):
```
$ nanoemoji --noclip_to_viewbox --color_format glyf_colr_1 0053.svg
```
<img width="264" alt="Screen Shot 2022-04-04 at 15 26 24" src="https://user-images.githubusercontent.com/6939968/161565562-1ab3549c-8422-43bd-b141-ec3aa2e0a98e.png">

resvg utility is making a bitmap of given pixel width/height, mapping the SVG's viewBox width/height onto the latter and clipping away what is outside. We may be able to retain the portion outside the viewbox by first stretching the viewbox to fit the union of all the paths' bounding boxes so we rasterize the entire outline, but then we have to think about how to reposition that bitmap in the font space. Which is already quite complicated, see #401

anthrotype:
a linked problem is the fact that the bitmap pixel width/height cannot exceed 255 because CBDT metrics use uint8, so even if we do support the viewbox overflow, we could still incur in the nastier integer overflow:

https://github.com/googlefonts/nanoemoji/pull/400#discussion_r841751567

rsheeter:
Ew. If we change the bitmap size our core computation of where the bitmap gets placed in font space breaks.

rsheeter:
I think we have to pre-process, "remember" the initial viewbox, alter the viewbox to include the stuff outside, rasterize that, and then use the initial viewbox when positioning the bitmap in font space. Tiresome but doable.

rsheeter:
This one from Amiri Quran Colored just isn't ever going to work well as a bitmap with these limits. I'm going to have to set the height in px to like ... 8.

[...]

PR #412, opened 2022-04-28
color_glyph: remove assertion that width > 0

anthrotype:
it's valid for fonts to have glyphs with zero advance width, e.g. combining marks

I stumbled on this while running `maximum_color` on a yet-to-be-published color font project, after exporting from Glyphs.app as an OT-SVG.

I wonder why the assertion was put that to begin with. Is it because when width is 0, a bitmap CBDT glyph would otherwise completely disappear (https://github.com/googlefonts/nanoemoji/issues/402)?


rsheeter:
> Is it because when width is 0, a bitmap CBDT glyph would otherwise completely disappear

I think this *was* true but iirc should no longer be. In any case, definitely a bad assert, good catch.

PR #414, opened 2022-04-29
Use precomputed part files for shape reuse

rsheeter:
Fixes #415. Fixes #313. Makes identifying and troubleshooting reuse failure far far easier as things in build/parts-merged.json with donor null after a build are the ones that failed affine_between, you can look to see if you agree with how things grouped, etc.

TODO list:

- [x] Merge #417 to mitigate excessive diff complexity :)
- [ ] Add an assert that the y-scale from combined parts => font space is 1 per https://github.com/googlefonts/nanoemoji/pull/417#pullrequestreview-1002620346
- [ ] Shapes seem good but gradients broken, see things like the svg for tests/clocks_picosvg.ttx or tests/smiley_cheeks_gradient_svg.ttx
- [ ] Add tests specifically targeting svg _glyph_groups
  * groups if an identical path appears in two color glyphs
  * groups if a path reused by transform appears
  * otherwise ... not
- [x] Rename ReuseInstruction to something like ReuseResult; XInstruction makes everyone think hints
- [ ] Check we can still build and render Noto Emoji & all demo fonts in color-fonts
  - [x] `nanoemoji ../svg/emoji_u1f6d2.svg` from https://github.com/googlefonts/noto-emoji/tree/colrv1/colrv1 fails
  - [ ] check clock emoji; they are getting significantly different ttx 
- [ ] Check that fonts didn't get bigger, ideally they got slightly smaller
  - [x] Noto Emoji colrv1, no flags is slightly smaller (2932768 => 2858124 bytes, -74644), probably because we are able to prefer to create smaller from larger instead of taking shapes in the order of encounter
  - [x] https://github.com/googlefonts/picosvg/pull/272 adds a further -34856
- [ ] Confirm https://github.com/rsheeter/color_aboard still works

PR #417, opened 2022-06-10
Generate part files for each svg, merge them, but don't yet use the result

rsheeter:
Split a comprehensible part of #414 out for merge: build part files, merge them, but don't actually use them for anything.

### DIFF
--- a/src/nanoemoji/write_combined_part_files.py
+++ b/src/nanoemoji/write_combined_part_files.py
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Generates a part file from 1 source, typically an svg file.
-"""
+"""Combines N part files to 1"""
 
 
 from absl import app
@@ -21,34 +20,33 @@ from absl import flags
 from nanoemoji.parts import ReusableParts
 from nanoemoji import util
 from pathlib import Path
-from picosvg.geometric_types import Rect
-from picosvg.svg import SVG
 
 
 FLAGS = flags.FLAGS
 
 
-flags.DEFINE_bool("compute_donors", False, "Whether to compute donors.")
-
-
 def main(argv):
-    if len(argv) != 2:
-        raise ValueError("Specify exactly one input")
+    input_files = util.expand_ninja_response_files(argv[1:])
 
-    view_box = Rect(0, 0, FLAGS.upem, FLAGS.upem)
-    parts = ReusableParts(view_box=view_box, reuse_tolerance=FLAGS.reuse_tolerance)
+    combined_parts = ReusableParts()
+    individual_parts = [ReusableParts.loadjson(Path(p)) for p in input_files]
+    if individual_parts:
+        combined_parts.version = util.only({p.version for p in individual_parts})
+        combined_parts.reuse_tolerance = util.only(
+            {p.reuse_tolerance for p in individual_parts}
+        )
+        combined_parts.view_box =  util.only(
+            {p.view_box for p in individual_parts}
+        )
 
-    svg = SVG.parse(Path(argv[1]))
-    parts.add(svg)
+    for parts in individual_parts:
+        combined_parts.add(parts)
 
-    if FLAGS.compute_donors:
-        parts.compute_donors()
+    combined_parts.compute_donors()  # precompute for later use
 
     with util.file_printer(FLAGS.output_file) as print:
-        print(parts.to_json())
+        print(combined_parts.to_json())
 
 
 if __name__ == "__main__":
-    flags.mark_flag_as_required("upem")
-    flags.mark_flag_as_required("reuse_tolerance")
     app.run(main)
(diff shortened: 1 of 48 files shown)


src/nanoemoji/write_combined_part_files.py
```diff
@@ -0,0 +1,50 @@
...
+    input_files = util.expand_ninja_response_files(argv[1:])
+
+    combined_parts = ReusableParts()
+    individual_parts = [ReusableParts.loadjson(Path(p)) for p in input_files]
+    if individual_parts:
+        combined_parts.version = util.only({p.version for p in individual_parts})
+        combined_parts.reuse_tolerance = util.only(
+            {p.reuse_tolerance for p in individual_parts}
+        )
+        combined_parts.view_box = util.only({p.view_box for p in individual_parts})
```
anthrotype, line 38: I think you did add support for merging parts having different viewboxes, didn't you? but here you require that all individual parts have the same viewbox

src/nanoemoji/write_part_file.py
```diff
@@ -12,37 +12,43 @@
...
+
 def main(argv):
-    parts = [ReuseableParts.load(Path(part_file)) for part_file in argv[1:]]
-    if not parts:
-        raise ValueError("Specify at least one input")
-    parts = reduce(lambda a, c: a.add(c), parts[1:], parts[0])
+    if len(argv) != 2:
+        raise ValueError("Specify exactly one input")
+
+    view_box = Rect(0, 0, FLAGS.upem, FLAGS.upem)
```
anthrotype, line 38: the relationship between an SVG viewBox and a font UPEM in nanoemoji is a bit more complex than that (remember how we scale to fit the viewBox.h to ascender+descender (not the UPEM proper) with the width centered or proportional).
I wonder, why did you feel you needed to add a `view_box` field to `ReusableParts` class? Why not just default to the unitary square (0, 0, 1, 1)? Doesn't reuse_tolerance already control how fine grained we want the reuse to be? What does a view_box (somewhat ambiguously assimilated to a --upem) add to this?

anthrotype: ah ok, the view_box on a part file acts as a reference point for the shape paths contained therein, just like the SVG.viewBox proper; so when you add a new part or merge two part files you can scale them accordingly so they are the same relative size.
Note the viewBox can vary in width for non-emoji fonts with proportional width; I believe you call Affine2D.rect_to_rect without preserveAspectRatio so you'd see their rectangular viewbox stretched to fit the nominal UPEM square, with shapes thus distorted..
Maybe that's ok, because when it's time to _use_ the combined reusable parts, I think you will have to remap again from this ReusableParts.view_box (currently set to (0, 0, upem, upem)) to the final, actual glyph area (delimited by ascender, descender, origin and advance width).

tests/conftest.py
```diff
@@ -45,6 +45,6 @@ def _cleanup_temporary_dirs(request):
         cleanup_temp_dirs()
     else:
         print(
-            f"NOT cleaning up {len(active_temp_dirs())} temp dirs to ease troubleshooting"
```
anthrotype, line 48: by the way, I think the `_cleanup_temporary_dirs` fixture should have a `scope="session"`, not "function" so they'll be torn down at the end of the session, not at the end of each test function, no?
https://docs.pytest.org/en/6.2.x/fixture.html#fixture-scopes

src/nanoemoji/write_combined_part_files.py
rsheeter: I did, but as all the individual parts are supposed to have `0 0 upem upem` boxes we should only get one here.

tests/conftest.py
rsheeter: Maybe I'm missing something obvious but ... why do I want them to last to the end of the session as opposed to being destroyed at the end of the test? - they shouldn't ever share across tests so we're done with them at the end of the test.

src/nanoemoji/write_part_file.py
rsheeter: I want all the parts for all glyphs to exist in the same space. Using a unit square would be lovely if scaling down to it and back up again didn't cause damage.

Idea was to stick the parts into the destination coordinate system to minimize damage from repeat transform. As you point out, perhaps we should do a square with asc-desc dimensions so the transform into the destination space has no scale. In fact, I should probably assert sx and sy are almost zero.

rsheeter: I changed it to use asc-desc. Mapping the shape back into font space isn't in this PR; I'll add a note to the todo list in #414.

rsheeter: I'll add a test for *not* butchering shapes when loading them into the part file, that's a good catch.

rsheeter: test_squares_stay_squares. Confirmed they were in fact getting mangled. I believe this is now fixed.

tests/conftest.py
rsheeter: Resolving for now, plz reopen if I need to change it

PR #419, opened 2022-06-10
Fix typos in maximum_color examples

justvanrossum:


anthrotype (approved):
good catch, thanks for trying out our tool!

rsheeter:
Oh ty, my bad

Issue #421, opened 2022-07-28
maximum_color is unhappy with > 1 palette


rsheeter:
In a clone of google/fonts:

```
maximum_color ofl/reemkufiink/ReemKufiInk-Regular.ttf

nanoemoji/colr_to_svg.py", line 410, in colr_to_svg
    assert len(ttfont["CPAL"].palettes) == 1, "We assume one palette"
AssertionError: We assume one palette
```

This is far less helpful than building a functional font with only one - presumably the first - palette.

anthrotype:
hm no good.

(LOL how the line just above that says `"""For testing only, don't use for real!"""`)

To fully support this we need https://github.com/googlefonts/nanoemoji/issues/422

> building a functional font with only one - presumably the first - palette.

anthrotype:
> building a functional font with only one - presumably the first - palette.

That's trivial, we can maybe do that as first round

rsheeter:
I like the idea of doing the trivial fix first :)

anthrotype:
if I apply the following patch, I can get past that assertion and the subsequent code will use `font["CPAL"].palettes[0]`

```diff
diff --git a/src/nanoemoji/colr_to_svg.py b/src/nanoemoji/colr_to_svg.py
index 9c514e9..4778558 100644
--- a/src/nanoemoji/colr_to_svg.py
+++ b/src/nanoemoji/colr_to_svg.py
@@ -407,7 +407,13 @@ def colr_to_svg(
     rounding_ndigits: Optional[int] = None,
 ) -> Dict[str, SVG]:
     """For testing only, don't use for real!"""
-    assert len(ttfont["CPAL"].palettes) == 1, "We assume one palette"
+    num_palettes = len(ttfont["CPAL"].palettes)
+    if num_palettes > 1:
+        logging.warning(
+            "Multiple CPAL palettes are not supported! Only using the first"
+        )
+    elif num_palettes < 1:
+        raise ValueError("No CPAL palettes found, at least one required")

     colr_version = ttfont["COLR"].version
     if colr_version == 0:
```

But... then I hit a new assertion error because some glyphs have zero advance width (which is normal and expected for glyphs used as combining marks).

```
W0729 15:25:50.514565 4340401536 colr_to_svg.py:412] Multiple CPAL palettes are not supported! Only using the first
Traceback (most recent call last):
...
[...]

Issue #405, opened 2022-04-11
Process currentColor properly in colr_to_svg.py


rsheeter:
https://github.com/googlefonts/nanoemoji/pull/400#pullrequestreview-936176907 identifies that black isn't always the current color. Fix it!

PR #423, opened 2022-07-29
take first in multi-palette CPAL; convert 0xFFFF palette index to SVG fill="currentColor"

anthrotype:
in an attempt to fix https://github.com/googlefonts/nanoemoji/issues/421 and allow ReekKufiInk be converted from COLRv1 to OT-SVG...

### DIFF
--- a/src/nanoemoji/colr_to_svg.py
+++ b/src/nanoemoji/colr_to_svg.py
@@ -108,7 +108,7 @@ def _first_cpal_palette(ttfont: ttLib.TTFont) -> List[C_P_A_L_.Color]:
 
 def _color(ttfont: ttLib.TTFont, palette_index, alpha=1.0) -> colors.Color:
     if palette_index == _FOREGROUND_COLOR_INDEX:
-        return colors.Color.fromstring("black")  # as good a guess as any
+        return colors.Color.fromstring("currentColor", alpha=alpha)
 
     # Only take the first palette for now:
     # https://github.com/googlefonts/nanoemoji/issues/421
--- a/src/nanoemoji/colr_to_svg.py
+++ b/src/nanoemoji/colr_to_svg.py
@@ -45,9 +45,10 @@ from picosvg.svg import SVG
 from picosvg.svg_meta import ntos
 from picosvg.svg_transform import Affine2D
 from fontTools import ttLib
+from fontTools.ttLib.tables import C_P_A_L_
 from picosvg.geometric_types import Point, Rect
 from lxml import etree
-from typing import Any, Callable, Dict, Iterable, Mapping, Optional, Tuple
+from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Tuple
 from fontTools.pens import transformPen
 from fontTools.ttLib.tables import otTables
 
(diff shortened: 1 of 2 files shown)


rsheeter:
I do wonder if the 0 width check shouldn't still exist somewhere for bitmaps only. Still seems like a good change. Can we add something like a very narrow subset of the arabic mark example to expose these issues in a test?

anthrotype:
i'll disentangle the issue about multiple CPAL palettes (this PR) from the one about zero-width glyphs, for which I am filing a separate PR https://github.com/googlefonts/nanoemoji/pull/424

anthrotype:
I rebased this on top of #424, so the latter should be merged first. This PR basically adds the last commit [6e8e43c](https://github.com/googlefonts/nanoemoji/pull/423/commits/6e8e43cf28c13c1312083d9b2c90728bc9c14a83)

src/nanoemoji/colr_to_svg.py
```diff
@@ -407,7 +399,13 @@ def colr_to_svg(
     rounding_ndigits: Optional[int] = None,
 ) -> Dict[str, SVG]:
     """For testing only, don't use for real!"""
-    assert len(ttfont["CPAL"].palettes) == 1, "We assume one palette"
+    num_palettes = len(ttfont["CPAL"].palettes)
+    if num_palettes > 1:
+        logging.warning(
+            "Multiple CPAL palettes are not supported! Only using the first"
+        )
+    elif num_palettes < 1:
```
rsheeter, line 407: Optional, control flow would be simpler if you just replace the original assert with > 0, must have at least one palette.

Hm. I'm tempted to suggest it should be OK to not have any palettes, then the only color you can use is foreground.

rsheeter (approved):

src/nanoemoji/colr_to_svg.py
anthrotype: > I'm tempted to suggest it should be OK to not have any palettes, then the only color you can use is foreground.

it took me reading the sentence three times to realise it was a joke

anthrotype: actually, you _are_ serious, you mean COLR without CPAL..

anthrotype: note to self, a CPAL the table still is required for COLR:

https://docs.microsoft.com/en-us/typography/opentype/spec/colr
> If the COLR table is present in a font but no CPAL table exists, then the COLR table is ignored.

but I suppose it could well contain no palettes if only foreground is used

rsheeter: > If the COLR table is present in a font but no CPAL table exists, then the COLR table is ignored.

I want to change that too, I think COLR without CPAL is fine. But we can leave it until COLRv2 to be consistent with the spec if you prefer.

anthrotype: ok, I'll keep a logging.warning for len(palettes) > 1, but move the check for CPAL presence and len(palettes) > 0 closer to where it's needed (inside `_color` function where it looks up a color in the CPAL palette) and will no longer assert that a non empty CPAL exists if all colors are foregrounds

anthrotype:
@rsheeter I also took the liberty to fix #405 with https://github.com/googlefonts/nanoemoji/pull/423/commits/0c77dfbec425b36bed631b2ff64bff605a355376 since I was touching those lines anyway and was a small fix

rsheeter:
Nice. Once this is merged/released you should maximum_color reem kufi ink on google/fonts :)

anthrotype:
> Once this is merged/released...

https://github.com/googlefonts/nanoemoji/releases/tag/v0.14.0

anthrotype:
> ...you should maximum_color reem kufi ink on google/fonts :)

FYI I'm travelling tomorrow so I won't be able to do that until next week (in cause you want to give that a try)

PR #428, opened 2022-08-31
[SVG] ensure colored .notdef stays as gid0 when reordering svg glyphs for reuse

### LINKED ISSUE #427: [maximum_color] colored .notdef produces invalid glyph order/cmap when COLR=>SVG
The [BungeeSpice](https://github.com/djrrb/Bungee/blob/master/fonts/Bungee_Color_Fonts/BungeeColor-Regular_colr_Windows.ttf-colrv1.ttf) font has a colored .notdef glyph.
When running maximum_color COLR=>SVG, we reorder glyphs such that color glyphs that share some shapes are placed in continuous ranges so we can use a single SVG document for each reuse group. We place all non-color glyphs at the front of the glyph order, followed by the contiguous color glyph ranges from the reuse groups.
Now, in this particular case this means that "space" (a non-color, since it's empty) gets shifted to the first glyph ID (0), and the colored .notdef is now in second position. This creates problems for cmap subtables which implicitly interpret gid0 to be the .notdef glyph. Effectively, the SPACE character 0x20 is lost and is no longer mapped to the "space" glyph after running through maximum_color on a font such as BungeeSpice that contains a colored .notdef glyph...
A font that doesn't contain a mapping for 0x20 SPACE character is undesirable, not just because we borked the original font which did contain one, but more generally because it triggers different fallback behavior across different environments.

The fix is to treat .notdef specially and ensure it does not get shuffled with the rest of the color glyphs when going from COLR=>SVG.

anthrotype:
Fixes #427

### DIFF
--- a/src/nanoemoji/svg.py
+++ b/src/nanoemoji/svg.py
@@ -127,11 +127,11 @@ def _glyph_groups(
 
     # If the color glyphs contain '.notdef', we want to keep it the first glyph, so
     # we must avoid grouping it with the rest, as they get reshuffled later on.
-    has_color_notdef = False
+    initial_glyphs = ()
     for color_glyph in color_glyphs:
         if color_glyph.ufo_glyph_name == ".notdef":
             assert color_glyph.glyph_id == 0
-            has_color_notdef = True
+            initial_glyphs = ((".notdef",),)
             break
 
     # ensure glyphs sharing shapes are in the same doc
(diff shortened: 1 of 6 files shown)


src/nanoemoji/svg.py
```diff
@@ -149,7 +165,12 @@ def _glyph_groups(
 
                 nth_paint_glyph += 1
 
-    return reuse_groups.sorted()
+    sorted_groups = reuse_groups.sorted()
+
+    if has_color_notdef:
+        return ((".notdef",),) + sorted_groups
+    else:
+        return sorted_groups
```
rsheeter, line 173: never liked this sort of branched return, wdyt about dropping the bool, storing a default-empty tuple of initial_glyphs, setting initial_glyphs to `(".notdef",)` above if there is one, and then here returning `return  (initial_glyphs,) + reuse_groups.sorted()`

rsheeter (approved):

src/nanoemoji/svg.py
anthrotype: OK I can change that. But why don't you like them?

PR #431, opened 2022-09-05
support OT-SVG color variables for CPAL palette entries

### LINKED ISSUE #422: Support OT-SVG with CSS var(--color0, ...) and color palettes from CPAL table
In addition to the inlined SVG colors (constant names "yellow" or hex #ABCDEF) which we do support, and the special `currentColor` keyword, which we don't yet (cf. #405), OT-SVG table also allows to define color variables (via `CSS var()` function) that reference color palettes defined in the CPAL table, thus allowing to change the colors by selecting among multiple color palettes (e.g. light vs dark, etc.).

We need to support this both ways: 1) in maximum_color when converting OT-SVG => COLR, and 2) when converting a COLR font with multiple CPAL palettes to OT-SVG (#421).

anthrotype:
Fixes #422 

~~Needs tests, don't merge yet.~~ Tests added

### DIFF
--- a/src/nanoemoji/colors.py
+++ b/src/nanoemoji/colors.py
@@ -200,7 +200,7 @@ class Color(Sequence):
 
     # the default color is optional but for now we require one for simplicity
     _COLOR_VARIABLE_RE: ClassVar[re.Pattern] = re.compile(
-        r"var\(\s*--color([0-9]+)\s*,\s*(#?\w+)\s*\)"
+        r"var\s*\(\s*--color([0-9]+)\s*,\s*(#?\w+)\s*\)"
     )
 
     def __getitem__(self, i):
--- a/src/nanoemoji/colors.py
+++ b/src/nanoemoji/colors.py
@@ -12,8 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import collections
-from typing import Optional, Sequence, Tuple
+import dataclasses
+import re
+from typing import ClassVar, Optional, Sequence, Tuple
 
 # See https://www.w3.org/TR/css-color-4/#named-colors
 # Chrome DevTools:
@@ -185,13 +186,46 @@ def color_name(rgb) -> Optional[str]:
     return None
 
 
-class Color(collections.namedtuple("Color", "red green blue alpha")):
-    _CURRENT_COLOR = (-1, -1, -1)
+@dataclasses.dataclass(frozen=True, order=True)
+class Color(Sequence):
+    red: int
+    green: int
+    blue: int
+    alpha: float = 1.0
+    index: Optional[int] = None
+
+    _CURRENT_COLOR: ClassVar[Tuple[int, int, int]] = (-1, -1, -1)
+
+    # the default color is optional but for now we require one for simplicity
+    _COLOR_VARIABLE_RE: ClassVar[re.Pattern] = re.compile(
+        r"var\(\s*--color([0-9]+)\s*,\s*(#?\w+)\s*\)"
+    )
+
+    def __getitem__(self, i):
+        fields = dataclasses.fields(self)
+        if isinstance(i, slice):
+            return tuple(getattr(self, f.name) for f in fields[i])
+        return getattr(self, fields[i].name)
+
+    def __len__(self):
+        return len(dataclasses.fields(self))
+
+    def _replace(self, **kwargs):
+        return dataclasses.replace(self, **kwargs)
 
     @classmethod
-    def fromstring(cls, s, alpha=1.0) -> "Color":
-        # https://www.w3.org/TR/css-color-4/#hex-notation
+    def fromstring(cls, s: str, alpha: float = 1.0) -> "Color":
         s = s.strip()
+
+        # CSS variables referencing palette entries as custom properties:
+        # https://docs.microsoft.com/en-us/typography/opentype/spec/svg#color-palettes
+        m = cls._COLOR_VARIABLE_RE.match(s)
+        if m:
+            palette_entry_index = int(m.group(1))
+            default_color = m.group(2)
+            tmp = cls.fromstring(default_color, alpha=alpha)
+            return tmp._replace(index=palette_entry_index)
+
         if s == "currentColor":
             # For the 'currentColor' special keyword, we return a sentinel value (with
             # negative invalid R G B values) that we'll convert to the 0xFFFF foreground
@@ -199,6 +233,7 @@ class Color(collections.namedtuple("Color", "red green blue alpha")):
             # https://docs.microsoft.com/en-us/typography/opentype/spec/SVG#colors
             return cls(*cls._CURRENT_COLOR, alpha=alpha)
         if s.startswith("#"):
+            # https://www.w3.org/TR/css-color-4/#hex-notation
             ss = s[1:]
             if len(ss) in (3, 4):
                 ss = "".join((s + s for s in ss))
@@ -242,6 +277,9 @@ class Color(collections.namedtuple("Color", "red green blue alpha")):
 
     def to_string(self) -> str:
         # A CSS or SVG friendly string
+        if self.index is not None:
+            return f"var(--color{self.index}, {self.default().to_string()})"
+
         if self.is_current_color():
             if self.alpha != 1.0:
                 raise ValueError("'currentColor' can't encode alpha != 1.0")
@@ -259,7 +297,12 @@ class Color(collections.namedtuple("Color", "red green blue alpha")):
     def is_current_color(self):
         return self[:3] == self._CURRENT_COLOR
 
+    def default(self) -> "Color":
+        if self.index is None:
+            return self
+        return self._replace(index=None)
+
     def palette_index(self, palette: Sequence["Color"]) -> int:
         if self.is_current_color():
             return 0xFFFF
-        return palette.index(self)
+        return palette.index(self.default())
--- a/src/nanoemoji/colors.py
+++ b/src/nanoemoji/colors.py
@@ -196,8 +196,6 @@ class Color(Sequence):
     alpha: float = 1.0
     index: Optional[int] = None
 
-    _CURRENT_COLOR: ClassVar[Tuple[int, int, int]] = (-1, -1, -1)
-
     # the default color is optional but for now we require one for simplicity
     _COLOR_VARIABLE_RE: ClassVar[re.Pattern] = re.compile(
         r"var\s*\(\s*--color([0-9]+)\s*,\s*(#?\w+)\s*\)"
@@ -233,7 +231,7 @@ class Color(Sequence):
             # negative invalid R G B values) that we'll convert to the 0xFFFF foreground
             # CPAL color palette index.
             # https://docs.microsoft.com/en-us/typography/opentype/spec/SVG#colors
-            return cls(*cls._CURRENT_COLOR, alpha=alpha)
+            return cls.current_color(alpha=alpha)
         if s.startswith("#"):
             # https://www.w3.org/TR/css-color-4/#hex-notation
             ss = s[1:]
@@ -296,8 +294,13 @@ class Color(Sequence):
                 string += f"{int(self.alpha * 255):02X}"
         return string
 
+    @classmethod
+    def current_color(cls, alpha=alpha) -> "Color":
+        # sentinel value for "currentColor" (text foreground color)
+        return cls(-1, -1, -1, alpha=alpha)
+
     def is_current_color(self):
-        return self[:3] == self._CURRENT_COLOR
+        return self[:3] == self.current_color()[:3]
 
     def default(self) -> "Color":
         if self.index is None:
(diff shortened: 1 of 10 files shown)


anthrotype:
@rsheeter PTAL. I think this is good enough to merge (it allows to build Nabla OT-SVG with multiple CPAL palettes from its COLRv1 variant).

src/nanoemoji/colors.py
```diff
@@ -185,20 +186,54 @@ def color_name(rgb) -> Optional[str]:
     return None
 
 
-class Color(collections.namedtuple("Color", "red green blue alpha")):
-    _CURRENT_COLOR = (-1, -1, -1)
+@dataclasses.dataclass(frozen=True, order=True)
+class Color(Sequence):
```
rsheeter, line 190: I agree we should move away from collections.namedtuple but ... why Sequence as opposed to NamedTuple from typing? - probably warrants a comment since it feels like a non-obvious choice

src/nanoemoji/colors.py
```diff
@@ -185,20 +186,54 @@ def color_name(rgb) -> Optional[str]:
...
+    red: int
+    green: int
+    blue: int
+    alpha: float = 1.0
+    index: Optional[int] = None
+
+    _CURRENT_COLOR: ClassVar[Tuple[int, int, int]] = (-1, -1, -1)
```
rsheeter, line 197: Shouldn't this be an instance of Color?

src/nanoemoji/colors.py
```diff
@@ -185,20 +186,54 @@ def color_name(rgb) -> Optional[str]:
...
+
+    # the default color is optional but for now we require one for simplicity
+    _COLOR_VARIABLE_RE: ClassVar[re.Pattern] = re.compile(
+        r"var\(\s*--color([0-9]+)\s*,\s*(#?\w+)\s*\)"
```
rsheeter, line 201: nit: I believe this will fail to match `var<whitespace>(`

src/nanoemoji/colors.py
```diff
@@ -185,20 +186,54 @@ def color_name(rgb) -> Optional[str]:
...
 
-class Color(collections.namedtuple("Color", "red green blue alpha")):
-    _CURRENT_COLOR = (-1, -1, -1)
+@dataclasses.dataclass(frozen=True, order=True)
+class Color(Sequence):
+    red: int
+    green: int
+    blue: int
+    alpha: float = 1.0
+    index: Optional[int] = None
```
rsheeter, line 195: suggest palette_index; just index isn't very obvious.

Needs a class comment to explain what it means to set or unset palette_index. For example, if I set a red value and a palette_index ... what does that mean?

src/nanoemoji/colors.py
```diff
@@ -185,20 +186,54 @@ def color_name(rgb) -> Optional[str]:
...
+    # the default color is optional but for now we require one for simplicity
+    _COLOR_VARIABLE_RE: ClassVar[re.Pattern] = re.compile(
+        r"var\(\s*--color([0-9]+)\s*,\s*(#?\w+)\s*\)"
+    )
+
+    def __getitem__(self, i):
+        fields = dataclasses.fields(self)
+        if isinstance(i, slice):
+            return tuple(getattr(self, f.name) for f in fields[i])
+        return getattr(self, fields[i].name)
```
rsheeter, line 208: This bit would go away if it was a named tuple I think?

src/nanoemoji/write_font.py
```diff
@@ -576,23 +576,42 @@ def _colr_ufo(
...
+        for c in chain.from_iterable(g.colors() for g in color_glyphs)
+        if not c.is_current_color()
     )
+    # Sort colors so the index into colors == index into CPAL palette.
+    # Try to keep the original palette entry index if present.
+    iter_colors_without_index = iter(sorted(c for c in colors if c.index is None))
+    colors_by_index = {}
+    for c in colors:
+        if c.index is not None:
+            color = c.default()
```
rsheeter, line 592: I don't find the behavior or name of default very intuitive. c.without_palette_index() or some such would at least tell me clearly what the intent here is.

src/nanoemoji/write_font.py
```diff
@@ -576,23 +576,42 @@ def _colr_ufo(
...
+            colors_by_index[c.index] = color
+    black = Color(0, 0, 0, 1.0)
+    colors = [
+        (
+            colors_by_index[i]
+            if i in colors_by_index
+            else next(iter_colors_without_index, black)
+        )
+        for i in range(max(colors_by_index, default=-1) + 1)
+    ] + list(iter_colors_without_index)
```
rsheeter, line 607: I found this a touch hard to follow. Suggested alternate implementation, which may simply prove I didn't follow at all:

```python

def _color_key(color):
  # sort indexed colors by their index, unindexed colors => end
  # all unindexed colors with the same rgb will group together
  if color.has_index:
    return (color.index,)
  return (len(colors), color[:3]) 

palette = []
for key, color in groupby(sorted(colors, key=_color_key), _color_key):
  color = tuple(color)
  if len(color) != 1:
    raise ValueError(f"Multiple colors are assigned {key}: {colors}")
  color = color[0]
  # if there are unused indices blow up
  if len(key) == 1:
    assert len(palette) == key[0], f"Color for index {key[0]} landed at index {len(palette)}"
  palette.append(color)
```

tests/colors_test.py
```diff
@@ -37,6 +37,9 @@
         ("#00A1DE\n", Color(0, 161, 222, 1.0)),
         # 'currentColor' is a special keyword
         ("currentColor", Color(-1, -1, -1, 1.0)),
+        # CSS variables for CPAL palette entry indices
+        ("var(--color0, red)", Color(255, 0, 0, 1.0, index=0)),
+        ("var(--color123, #ABCDEF)", Color(0xAB, 0xCD, 0xEF, 1.0, index=123)),
```
rsheeter, line 42: Suggest adding one or two with weird spacing

`var  (  --c1\t,...etc`

tests/colors_test.py
```diff
@@ -58,7 +61,28 @@ def test_color_fromstring(color_string, expected_color):
...
+def test_color_like_namedtuple():
+    color = Color(0x00, 0x11, 0x22, 1.0, index=4)
+
+    assert color.red == color[0] == 0x00
+    assert color.green == color[1] == 0x11
+    assert color.blue == color[2] == 0x22
+    assert color.alpha == color[3] == 1.0
+    assert color.index == color[4] == 4
+
+    assert len(color) == 5
```
rsheeter, line 82: I'm again bothered that it isn't *actually* a namedtuple

rsheeter:
How much does the size of the SVG table for Nabla change with/without this? - not suggesting we block the PR, just an interesting datapoint

rsheeter (approved):
LGTM.

Nit: color not being an actual typing.NamedTuple throws me, further comments/questions on that inline.

anthrotype:
Re: namedtuple vs dataclass, i hinted at the reason in the commit message, I can can add a comment maybe:
https://github.com/googlefonts/nanoemoji/pull/431/commits/043bcdfceacb594d5dc0121af94c85bb3079b399

In general I think the dataclass interface is much superior to namedtuple, but if you really prefer the latter I can change it back to that (and do away with Class at which namedtuple can't support).

I'll address the rest of the comments next week when I'm back, thanks for the review

src/nanoemoji/colors.py
anthrotype: Color class hasn't completely been defined yet, can't make an instance of it from within the class scope (btw this code preexists my PR)

src/nanoemoji/colors.py
anthrotype: Sequence is the closest collections.abc to a tuple, i wanted to keep the tuple-like interface but use a dataclass instead (can't have both a dataclass and namedtuple, they don't play nicely togheter)

src/nanoemoji/colors.py
anthrotype: Good catch will fix, though unlikely in real world svgs

src/nanoemoji/colors.py
anthrotype: it means var(--color{self.index}, red}, like in OT-SVG color variables. Red is the default or fallback value for an implementation that does not support fetching the corresponding CPAL palette entry at given index.
I didn't use palette_index because I thought that might get confused for the index into the array of palettes, but it's actually the palette _entry_ index (which is too verbose but maybe it's clearer). Perhaps I'll change it as you suggested or add a comment

src/nanoemoji/colors.py
anthrotype: Yes, it would

src/nanoemoji/write_font.py
anthrotype: Yeah. I was thinking along the lines of "the default value if CPAL not supported".

src/nanoemoji/write_font.py
anthrotype: The intent of that a bit convoluted code was to keep the color indices of the CPAL we generate as originally specified by the var(--color{index}) variables in the input SVG, but also sort the ones that do not specify an explicit index, placing the in the empty slots in between those with explicit index and any remaining ones appended at the end.

src/nanoemoji/colors.py
rsheeter: I'd vote for long name over one that requires a comment

src/nanoemoji/colors.py
rsheeter: Why do you specifically want/need a dataclass? I see mildly ugly code to make it pretend and I wasn't clear on the benefit

src/nanoemoji/colors.py
rsheeter: There must be a way to make it an actual instance? Make a method akin to Affine2D.identity()? (nice for consistency too)

src/nanoemoji/write_font.py
rsheeter: It's the default value part that throws me I think. In what sense does an rgba color have a default?

anthrotype: It's a color var() that can have a default (which in turn is a color). Initially I went for a separate class for color variables, but then I would have to basically rewrite all the places where we accept Color to also accept a Color variables. It was just much easier to change Color to optionally carry a (palette entry) index.

src/nanoemoji/colors.py
rsheeter: (I don't especially object, just curious)

src/nanoemoji/write_font.py
rsheeter: Ah, so my first attempt above doesn't do the part about filling in holes. #432 shows my second attempt as I still find the current code a bit hard to follow. Feel free to use or ignore.

anthrotype:
> How much does the size of the SVG table for Nabla change with/without this?

the total font file size increased by 45KB, from 1,598,000 bytes to 1,643,616 bytes (of course, all due to the more verbose SVG table)

anthrotype:
but the WOFF2 version is only 2.5KB larger, going from 185,004 to 187,540 bytes

src/nanoemoji/write_font.py
anthrotype: sorry but I think your code is even more convoluted than mine, perhaps it's a matter of style/opinions..

rsheeter:
> the dataclass interface is much superior to namedtuple

How so?

> the total font file size increased by 45KB, from 1,598,000 bytes to 1,643,616 bytes (of course, all due to the more verbose SVG table)

That seems like 45kb well spent to me :)

anthrotype:
> the dataclass interface is much superior to namedtuple

I already gave up, I'm changing it back to nanedtuple, life is too short

src/nanoemoji/write_font.py
rsheeter: For me at least when reading the code a name that directly says what it does (c.rgba_only, c.without_palette_index, etc) would be clearer. That said, I'm fine if you want to leave the name default.

src/nanoemoji/write_font.py
rsheeter: > your code is even more convoluted than mine

I was a little worried that might be the case lol

rsheeter:
> https://peps.python.org/pep-0557/#why-not-just-use-namedtuple

Ty, I had not seen that. That seems sufficient to have a strong preference for dataclasses indeed. ...now I wonder if we have other namedtuples we should remove!

anthrotype:
ha.. so shall I revert 588fdae now?

rsheeter:
> ha.. so shall I revert https://github.com/googlefonts/nanoemoji/commit/588fdaeaf20e95560f1f51a060352671f988acf0 now?

Yes please. Sorry for the delay and thank you - for the many'th time - for taking the time to explain python basics to me :D

src/nanoemoji/colors.py
anthrotype: palette_index() is already the name of the method that looks up a color in a list of Colors and returns the index (or 0xFFFF for currentColor). I'll rename the latter method `index_from(self, palette: Sequence["Color"]) -> int` so that I can rename the field `palette_index` as you suggested.

src/nanoemoji/write_font.py
anthrotype: changed to without_palette_index()

PR #432, opened 2022-09-09
Trying to simplify color sort

rsheeter:


src/nanoemoji/write_font.py
```diff
@@ -573,45 +573,45 @@ def _colr_ufo(
...
+        return (cpal_slots + 1,) + c[:4]
+
+    # Push colors into CPAL, either at their index or at the next open slot
+    cpal_colors = deque(sorted(cpal_colors, key=_color_sort_key))
+    colors = [black] * cpal_slots
+    for i in range(cpal_slots):
+        if i == cpal_colors[0].index:
+            colors[i] = cpal_colors.popleft().default()  # TODO rename default
+        elif cpal_colors[-1].index is None:
+            colors[i] = cpal_colors.pop()
```
anthrotype, line 611: you're popping unindexed colors (that you previously sorted by increasing RGBA value) from the _end_ of the deque, so they will be stored in the cpal in reverse order. Even though it shouldn't matter for these in which order they are stored, previous code would store them in increasing RGBA value (not decreasing).

src/nanoemoji/write_font.py
```diff
@@ -573,45 +573,45 @@ def _colr_ufo(
...
-    colors_by_index = {}
-    for c in colors:
+
+    # Chrome 98 doesn't like when COLRv1 font has empty CPAL palette so make sure we have at least one
+    # TODO(anthrotype): File a bug and remove hack once the bug is fixed upstream
+    if not cpal_colors:
+        cpal_colors = {black}
+
+    # We need enough slots for all the colors, or to reach the max index, whichever is greater
+    cpal_slots = max(len(cpal_colors), max(c.index or -1 for c in cpal_colors))
```
anthrotype, line 595: indices start counting at 0, so to compute the length of the final list you need to add +1 to the maximum index

```suggestion
    cpal_slots = max(len(cpal_colors), max(c.index or -1 for c in cpal_colors) + 1)
```

src/nanoemoji/write_font.py
rsheeter: Fixed

PR #434, opened 2022-09-20
maximum_color: skip svg table generation for colr v0 fonts

m4rc1e:
For fonts released on Google Fonts, we don't want to run maximum_color on colr v0 fonts since they're well supported in modern browsers.

Feel free to close this if you feel it's too restrictive. I can always move this functionality to our own gftools builder so it will only work on fonts destined for GF.

### DIFF
--- a/src/nanoemoji/maximum_color.py
+++ b/src/nanoemoji/maximum_color.py
@@ -440,10 +440,16 @@ def _run(argv):
         input_file.resolve() != (build_dir() / final_output).resolve()
     ), "In == Out is bad"
 
+    color_table = _vector_color_table(font)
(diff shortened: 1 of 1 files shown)


m4rc1e:
No doubt the tests are probably going to explode here.

anthrotype:
I don't think I want to do that. The tool supports this, there may be other valid reasons for one wanting to make an OT-SVG table from a COLRv0 font nonetheless (e.g. some app out there may support only the former). It's up to the user to decide whether to do it or not.

m4rc1e:
Ok cool. I'll move this into our own builder.

rsheeter:
+1 to doing this in GF specific code.

Issue #426, opened 2022-08-26
colr_to_svg: Support Format 5 (PaintVarLinearGradient)


m4rc1e:
Just tried running the maximum_color tool on [Foldit](https://github.com/SophiaDesign/Foldit/releases/tag/1.000) and got the following traceback:

```PYTHON
Traceback (most recent call last):
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/Users/marcfoley/Type/nanoemoji/src/nanoemoji/generate_svgs_from_colr.py", line 63, in <module>
    app.run(main)
  File "/Users/marcfoley/Type/tools/venv/lib/python3.10/site-packages/absl/app.py", line 312, in run
    _run_main(main, args)
  File "/Users/marcfoley/Type/tools/venv/lib/python3.10/site-packages/absl/app.py", line 258, in _run_main
    sys.exit(main(argv))
  File "/Users/marcfoley/Type/nanoemoji/src/nanoemoji/generate_svgs_from_colr.py", line 54, in main
    for glyph_name, svg in colr_to_svg(lambda gn: _view_box(font, gn), font).items():
  File "/Users/marcfoley/Type/nanoemoji/src/nanoemoji/colr_to_svg.py", line 428, in colr_to_svg
    svgs = _colr_v1_to_svgs(view_box_callback, ttfont)
  File "/Users/marcfoley/Type/nanoemoji/src/nanoemoji/colr_to_svg.py", line 403, in _colr_v1_to_svgs
    return {
  File "/Users/marcfoley/Type/nanoemoji/src/nanoemoji/colr_to_svg.py", line 406, in <dictcomp>
    _colr_v1_glyph_to_svg(ttfont, glyph_set, view_box_callback, g)
  File "/Users/marcfoley/Type/nanoemoji/src/nanoemoji/colr_to_svg.py", line 363, in _colr_v1_glyph_to_svg
    _colr_v1_paint_to_svg(
  File "/Users/marcfoley/Type/nanoemoji/src/nanoemoji/colr_to_svg.py", line 283, in _colr_v1_paint_to_svg
    descend(parent_el, child_paint)
[...]

PR #439, opened 2022-10-14
[maximum_color] Ignore variation data when converting COLRv1=>SVG

anthrotype:
Fixes https://github.com/googlefonts/nanoemoji/issues/426

### DIFF
--- a/src/nanoemoji/colr_to_svg.py
+++ b/src/nanoemoji/colr_to_svg.py
@@ -125,6 +125,19 @@ def _color(ttfont: ttLib.TTFont, palette_index, alpha=1.0) -> colors.Color:
     )
 
 
+# Map PaintVarXxx.Format to its corresponding non-var Format
+_PAINT_VAR_MAPPING = {
+    v: otTables.PaintFormat[k.replace("PaintVar", "Paint")]
+    for k, v in otTables.PaintFormat.__members__.items()
+    if k.startswith("PaintVar")
+}
+
+
+def _static_paint_format(fmt: int) -> otTables.PaintFormat:
+    fmt = otTables.PaintFormat(fmt)
+    return _PAINT_VAR_MAPPING.get(fmt, fmt)
+
+
 def _gradient_paint(ttfont: ttLib.TTFont, ot_paint: otTables.Paint) -> _GradientPaint:
     stops = tuple(
         ColorStop(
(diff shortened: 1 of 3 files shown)


m4rc1e:
Looks good on Safari

Screenshot 2022-10-14 at 14 12 07

*Foldit using this pr*

Screenshot 2022-10-14 at 14 13 01

*Foldit currently in google/fonts*

m4rc1e:
LGTM but I'm not a maintainer. I like the mapping approach you've used.

anthrotype:
Hm the spacing looks suspiciously bad in that screenshot.. maybe safari is applying the metrics of the regular but the SVG has the outlines of the thin?

m4rc1e:
I'm just checking to see if the testing site itself is to blame.

anthrotype:
It's not the site, it's the font that advertises itself as variable (has fvar, HVAR, etc.) but then the SVG outlines are not. Honestly I think it doesn't make much sense to have hybrid COLRv1+SVG when the COLR table contains variations, since there can't be a 1:1 translation between the two; besides, leaving variable tables alongside static SVG leads to issues like the one above. Only static COLRv1 font can be meaninfully converted to OT-SVG.
Once we have support for instancing COLRv1 in fonttools https://github.com/fonttools/fonttools/issues/2699 one could, say, cut an instance of the Regular or any other instance, and _then_ pass that through maximum_color to obtain the hybrid COLRv1+SVG.

anthrotype:
yes, that's expected. The font has HVAR, which varies along wght axis. The SVG doesn't move, the rest does..

anthrotype:
See discussion https://github.com/googlefonts/nanoemoji/issues/426#issuecomment-1228414249, in particular Rod's last reply

>> I believe (I hope) we currently subset away all the variation data when we serve these fonts to browsers that only understand OT-SVG but not COLR

>> We do not. The default request for a family is for a static instance, but if you ask for variation to be included we'll merrily do it.

anthrotype:
basically, the only way to currently make an OT-SVG font for Foldit-VF is to
1) run maximum_color on it
2) strip it of the variation data (instantiate it at the default location), 
3) drop the variable COLR table altogether

this will give you the Foldit-Thin.ttf (and only that one) with an SVG instead of a COLR table.

m4rc1e:
Seems better to just leave the font as is. At least that way Safari users will get a font without any color which imo is better than just the thin weight.

anthrotype:
maybe we need a separate tool (colr_instancer?) that takes a variable COLRv1 and spits out static OT-SVG fonts (for all named instances or specific selection). But that would first require implementing CORLv1 instancing in fontTools https://github.com/fonttools/fonttools/issues/2699

vv-monsalve:
> Seems better to just leave the font as is. At least that way Safari users will get a font without any color which imo is better than just the thin weight. If we leave as is, maximum_color could raise a warning for this issue.

From the above tests, I'd support this idea. Otherwise, it would break the whole idea of the font: serving a font that goes from thin-ultracondensed to black-extraexpanded with gradients that seems consistent through the entire range

Issue #435, opened 2022-09-21
KeyError: 'glyf' when trying to convert Gilbert font to COLR


Findus23:
Hi, 

I wanted to convert the Gilbert font from https://www.typewithpride.com/ to COLRv1 so that it also works in Chrome.

The default file name of `Gilbert-Color Bold Preview5.otf` seems to fail because of the spaces in the name.

But even after renaming the file, I can't seem to get it to work with `maximum_color /path/to/test.otf`:
```
I0921 12:00:26.850657 139996433276864 maximum_color.py:449] Generating build.ninja
I0921 12:00:27.044305 139996433276864 ninja.py:113] ninja -C /home/lukas/tmp/nanoemoji/build
ninja: Entering directory `/home/lukas/tmp/nanoemoji/build'
[904/906] /home/lukas/.virtualenvs/te...tput_file MergeSource.glyf_colr_1.ttf
W0921 12:01:18.666877 140660174471104 glyph_reuse.py:69] affine_between failed: M505.3699951171875,289.093994140625 L384.42999267578125,289.093994140625 L384.42999267578125,0.1820068359375 L505.3699951171875,0.1820068359375 L505.3699951171875,289.093994140625 M505.3699951171875,698.9450073242188 L384.42999267578125,698.9450073242188 L384.42999267578125,410.03399658203125 L505.3699951171875,410.03399658203125 L505.3699951171875,698.9450073242188 M160.9600067138672,288.7039794921875 L40.02000045776367,288.7039794921875 L40.02000045776367,-0.59100341796875 L160.9600067138672,-0.59100341796875 L160.9600067138672,288.7039794921875 M160.9600067138672,698.9249877929688 L40.02000045776367,698.9249877929688 L40.02000045776367,410.05401611328125 L160.9600067138672,410.05401611328125 L160.9600067138672,698.9249877929688 
[ some more lines like this]
I0921 12:01:25.416941 140660174471104 __init__.py:47] Pre-processing glyphs
I0921 12:01:25.456688 140660174471104 base.py:178] Running DecomposeComponentsFilter on AnEmojiFamily-Regular
I0921 12:01:25.457893 140660174471104 base.py:178] Running CubicToQuadraticFilter on AnEmojiFamily-Regular
I0921 12:01:25.530013 140660174471104 cubicToQuadratic.py:46] New spline lengths: 1: 844, 2: 796, 3: 372, 4: 30, 5: 9
[...]

PR #442, opened 2022-11-18
support input files containing spaces

anthrotype:
Part of fixing #435

### DIFF
--- a/tests/util_test.py
+++ b/tests/util_test.py
@@ -1,0 +1,54 @@
+# Copyright 2022 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import sys
+
+from nanoemoji.util import shell_quote, shell_split
+
+import pytest
+
+
+# Source:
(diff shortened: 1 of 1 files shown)


anthrotype:
The problem here was that ninja user-defined build variables (e.g. `--target_font $target_font` in `glue_together` rule) don't get shell-quoted automatically when they are expanded, only `$in` and `$out` are (hence the latter at least are fine when inputs/outputs contain spaces).
If the other variables happen to refer to path with spaces or other special characters, ninja passes them through unchanged and the underlying shell split the command line arguments incorrectly.
So we need to enquote them following the shell's own rules, which differ between POSIX and Windows, as usual.
Unfortunately the stdlib's `shlex` module's quote() and split() commands don't work well with Windows, so I had to resort to a third-party module called https://github.com/smoofra/mslex to accomplish this (it's only installed if sys.platform=="win32").
Now, `subprocess` also has a similar method called `list2cmdline` which follows Windows MS C runtime rules, but.. it's private and undocumented, and I also need the reverse of that in order to split a str into shell-compatible list of arguments (we need to split inside `nanoemoji.util.expand_ninja_response_files` to emulate the `@`-prefixed so-called response files used by ninja to work around maximum-args limits).

rsheeter:
https://github.com/smoofra/mslex looks like a tiny dead one-man project. I don't think this is a good dependency. Since Python can do things like run windows processes I'd have thought it *must* have this built-in but if not perhaps we should just write our own little function?

anthrotype:
ok, I removed the dependency on mslex, and went back to using subprocess.list2cmdline for shell quoting on Windows. Whereas for splitting a command line string to list of args, I am now using ctypes to wrap a call into Windows' own `CommandLineToArgvW` method (which works only on Windows of course, but that's ok). I copied the tests from cpyton's test_subprocess.py to confirm that both quoting and splitting back do roundtrip as intended on Windows.

rsheeter:
I'm mildly amazed where this has taken us haha.

rsheeter (approved):
LGTM

PR #448, opened 2023-03-14
take the min(sx,sy) when decomposing scale transform to minimise OverflowErrors

anthrotype:
WIP, will add test soon

### DIFF
--- a/src/nanoemoji/paint.py
+++ b/src/nanoemoji/paint.py
@@ -282,7 +282,9 @@ class PaintLinearGradient(Paint):
 
 def _decompose_uniform_transform(transform: Affine2D) -> Tuple[Affine2D, Affine2D]:
     scale, remaining_transform = transform.decompose_scale()
-    s = max(*scale.getscale())
+    # we take the minimum of (sx, sy) to try to minimize the risk of OverflowErrors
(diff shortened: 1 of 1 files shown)


src/nanoemoji/paint.py
```diff
@@ -282,7 +282,9 @@ def round(self, ndigits: int) -> "PaintLinearGradient":
 
 def _decompose_uniform_transform(transform: Affine2D) -> Tuple[Affine2D, Affine2D]:
     scale, remaining_transform = transform.decompose_scale()
-    s = max(*scale.getscale())
+    # we take the minimum of (sx, sy) to try to minimize the risk of OverflowErrors
+    # when encoding these to F16Dot16
+    s = min(*scale.getscale())
```
rsheeter, line 287: I immediately wonder if we should try both?

rsheeter:
Capturing @anthrotype's idea from this morning, we should make test inputs with big upem to exacerbate the problem

rsheeter:
Thinking "aloud" I wonder if SVD on the transform would make it easier to assemble a stack of transforms that work ... e.g. SVD, then search for parts you can put back together to minimize the # of paints you end up with

PR #452, opened 2023-05-21
Switch from deprecated "toml" library

### LINKED ISSUE #451: Switch from deprecated `toml` library
The replacement is Python's 3.11 `tomllib` module and `tomli` package for older versions.

CyberTailor:
Fixes #451

### DIFF
--- a/src/nanoemoji/config.py
+++ b/src/nanoemoji/config.py
@@ -20,9 +20,9 @@ except ImportError:
     import importlib_resources as resources  # pytype: disable=import-error
 
 import itertools
+import tomlkit
 from pathlib import Path
 from picosvg.svg_transform import Affine2D
-import toml
 from typing import Any, Iterable, MutableMapping, NamedTuple, Optional, Tuple, Sequence
 
 from nanoemoji import util
@@ -247,6 +247,20 @@ class FontConfig(NamedTuple):
         raise ValueError("Must have a default master")
 
 
+def toml_strip(data):
+    """
+    Workaround for:
+    https://github.com/sdispater/tomlkit/issues/240
+    """
+    new_data = {}
+    for key, val in data.items():
+        if isinstance(val, dict):
+            val = toml_strip(val)
+        if val is not None:
+            new_data[key] = val
+    return new_data
+
+
 def write(dest: Path, config: FontConfig):
     toml_cfg = {
         "family": config.family,
@@ -288,7 +302,8 @@ def write(dest: Path, config: FontConfig):
             for m in config.masters
         },
     }
-    dest.write_text(toml.dumps(toml_cfg))
+    with open(dest, "w") as toml_file:
+        tomlkit.dump(toml_strip(toml_cfg), toml_file)
 
 
 def _resolve_config(
@@ -296,9 +311,11 @@ def _resolve_config(
 ) -> Tuple[Optional[Path], MutableMapping[str, Any]]:
     if config_file is None:
         with resources.path("nanoemoji.data", _DEFAULT_CONFIG_FILE) as config_file:
-            # no config_dir in this context; bad input if we need it
-            return None, toml.load(config_file)
-    return config_file.parent, toml.load(config_file)
+            with open(config_file) as file:
+                # no config_dir in this context; bad input if we need it
+                return None, tomlkit.load(file)
+    with open(config_file) as file:
+        return config_file.parent, tomlkit.load(file)
 
 
 def _resolve_src(relative_base: Optional[Path], src: str) -> Iterable[Path]:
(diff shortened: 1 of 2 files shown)


rsheeter:
Ty for the PR. Apologies for the inconvenience but would you be willing to sign the CLA? - we can't accept otherwise.

CyberTailor:
I've signed it

CyberTailor:
https://github.com/googlefonts/nanoemoji/pull/452/checks?check_run_id=13637748132

anthrotype:
thanks. Can you please run black auto-formatter so that the lint CI job passes as well?

https://github.com/googlefonts/nanoemoji/actions/runs/5036873831/jobs/9036557954?pr=452

anthrotype:
@CyberTailor oh, I see the lint job failing has nothing to do with your specific PR, it's just a new black version that changed the way it wants to format things (and we keep deps un-pinned so this can happen /cc @rsheeter). You can ignore this, we can fix up later on after merge.

src/nanoemoji/config.py
```diff
@@ -297,8 +298,8 @@ def _resolve_config(
     if config_file is None:
         with resources.path("nanoemoji.data", _DEFAULT_CONFIG_FILE) as config_file:
             # no config_dir in this context; bad input if we need it
-            return None, toml.load(config_file)
-    return config_file.parent, toml.load(config_file)
+            return None, tomlkit.load(open(config_file))
```
anthrotype, line 301: ideally we should use `with open(...)` statement and let the file be closed on exit

src/nanoemoji/config.py
```diff
@@ -288,7 +288,8 @@ def write(dest: Path, config: FontConfig):
             for m in config.masters
         },
     }
-    dest.write_text(toml.dumps(toml_cfg))
+    toml_cfg = {key: val for key, val in toml_cfg.items() if val is not None}
```
anthrotype, line 291: why is None special now? is that a know difference between the old and new toml libraries?

src/nanoemoji/config.py
CyberTailor: Fixed

src/nanoemoji/config.py
CyberTailor: Yes, tomlkit raises an error on None values: https://github.com/sdispater/tomlkit/issues/240

CyberTailor: Added a comment for that workaround

src/nanoemoji/config.py
```diff
@@ -288,17 +303,19 @@ def write(dest: Path, config: FontConfig):
...
 def _resolve_config(
     config_file: Optional[Path] = None,
 ) -> Tuple[Optional[Path], MutableMapping[str, Any]]:
     if config_file is None:
-        with resources.path("nanoemoji.data", _DEFAULT_CONFIG_FILE) as config_file:
-            # no config_dir in this context; bad input if we need it
-            return None, toml.load(config_file)
-    return config_file.parent, toml.load(config_file)
+        # no config_dir in this context; bad input if we need it
+        toml_cfg = resources.files(nanoemoji.data).joinpath(_DEFAULT_CONFIG_FILE).read_text()
```
anthrotype, line 315: I see `importlib.resources.path` is now deprecated and they want us to use `filez` instead, that's ok. Does't that also support passing the package name as a string? "nanoemoji.data" instead of `import nanoemoji.data` that you did above? I would prefer the string if possible, thanks

src/nanoemoji/config.py
```diff
@@ -247,6 +248,20 @@ def default(self) -> MasterConfig:
         raise ValueError("Must have a default master")
 
 
+def toml_strip(data):
+    """
+    Workaround for:
+    https://github.com/sdispater/tomlkit/issues/240
```
anthrotype, line 254: ah! I actually didn't know toml didn't support Nones 👍

src/nanoemoji/config.py
CyberTailor: Fixed

anthrotype:
python 3.7 complains importlib.resources has no attribute "files"

https://github.com/googlefonts/nanoemoji/actions/runs/5044764986/jobs/9048275533?pr=452#step:5:71

do you mind not reverting that line change for now use the old deprecated method, we may want to drop 3.7 support at some point but maybe not immediately in this PR.

anthrotype:
I see that `files` method is 3.9+ only, we do want to support 3.8 for a bit longer. I would rather we keep the deprecation warning (or if we really don't want to see that, only use the newer API for 3.9+)

CyberTailor:
> do you mind not reverting that line change for now use the old deprecated method, we may want to drop 3.7 support at some point but maybe not immediately in this PR.

Reverted.

anthrotype:
thank you for the contribution and for the patience :)

PR #475, opened 2025-04-22
Expose and fix bug with production of high res sbix fonts

rsheeter:
For some definition of highres

Built on #476, merge that first


cmyr (approved):
I'm not familiar with this codebase but this all seems sane to me?

PR #480, opened 2025-06-03
Not a test

rsheeter:
Apparently this became a test and now blocks ci

### DIFF
--- a/tests/test_helper.py
+++ b/tests/test_helper.py
@@ -34,6 +34,7 @@ import shutil
 import tempfile
 
 
+@pytest.mark.skip(reason="It's not a test!!")
 def test_data_dir() -> Path:
     return Path(__file__).parent
 


rsheeter:
This is a long standing function that has suddenly become a release
blocking problem. I'm not quite sure what triggered it - maybe returning
data from tests used to be fine? - but I'd like to be able to release again.

anthrotype:
If I `pip install pytest==8.3.5` it works as usual, but as soon as I upgrade to the latest `pytest==8.4.0` then this issue is triggered.

it appears to be some new behavior, which can be disabled by setting `collect_imported_tests = false` in pytest config (e.g. setup.cfg):

https://docs.pytest.org/en/stable/reference/reference.html#confval-collect_imported_tests
https://docs.pytest.org/en/stable/changelog.html#pytest-8-4-0-2025-06-02

I think since it's just a single helper function, we can simply rename it to something that is unambiguously not a test

PR #481, opened 2025-06-03
[test_helper] rename helper function to circumvent pytest's overly greedy test discovery

anthrotype:
Meant to fix https://github.com/googlefonts/nanoemoji/pull/480

Something must have changed in pytest 8.4.0 which makes the test collection/discovery treat this little test helper function named 'test_data_dir' as a test function (pytest errors because it expects some asserts), which it is cleary not.

We make it a constant (Path objects are immutable) to workaround this

### DIFF
--- a/tests/config_test.py
+++ b/tests/config_test.py
@@ -15,7 +15,7 @@
 from nanoemoji import config
 from pathlib import Path
 import pytest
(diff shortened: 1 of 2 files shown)


rsheeter:
Now the test data dir is ANGRY WITH ME. But I'll take it lol.

Issue #472, opened 2024-08-20
consider always setting explicit 'fill'  even for default black paths as workaround for Firefox bug


anthrotype:
looks like Firefox renders implicitly black SVG paths with the current text color when they have no `fill` attribute, see:
https://github.com/googlefonts/noto-emoji/issues/480

this is where nanoemoji omits setting the `fill` attribute when its value is 'black' which is the default according to SVG spec:

https://github.com/googlefonts/nanoemoji/blob/fbd1a35916c2f22804a6139992d07e74480a9209/src/nanoemoji/svg.py#L191-L192

PR #488, opened 2025-10-01
colr_to_svg: always emit fill="black" to workaround FireFox issue 

anthrotype:
Despite fill="black" being the [default](https://www.w3.org/TR/SVG2/painting.html#SpecifyingFillPaint) initial value for the `fill` attribute in SVG, FireFox seems to be applying the current text color to those shapes that omit the fill attribute for a color font with SVG table.

If we always emit fill="black" explicitly instead of omitting it, the issue should go away and black shapes should stay black, no matter the current text color.

The SVG table size might increase slightly.

https://github.com/googlefonts/noto-emoji/issues/480
https://bugzilla.mozilla.org/show_bug.cgi?id=1908302

### DIFF
--- a/src/nanoemoji/svg.py
+++ b/src/nanoemoji/svg.py
@@ -188,6 +188,10 @@ def _apply_solid_paint(el: etree.Element, paint: PaintSolid):
         assert paint.color.opaque() == Color.fromstring(
             "black"
         ), "Unexpected color choice"
+    # Always set fill explicitly, even for black (the SVG default).
+    # This works around a Firefox rendering bug:
+    # https://github.com/googlefonts/noto-emoji/issues/480
+    # https://bugzilla.mozilla.org/show_bug.cgi?id=1908302
     el.attrib["fill"] = paint.color.opaque().to_string()
     if paint.color.alpha != 1.0:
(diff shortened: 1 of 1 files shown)


anthrotype:
Will fix https://github.com/googlefonts/nanoemoji/issues/472

tests/group_opacity_reuse_picosvg.ttx
```diff
@@ -61,8 +61,8 @@
     <svgDoc endGlyphID="2" startGlyphID="2">
       <![CDATA[<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1">
   <g id="glyph2" transform="matrix(0.781 0 0 0.781 0 -100)">
-    <path d="M60,10 L80,10 L80,60 L60,60 L60,10 Z" id="e000.0"/>
-    <g opacity="0.6">
+    <path d="M60,10 L80,10 L80,60 L60,60 L60,10 Z" fill="black" id="e000.0"/>
+    <g fill="black" opacity="0.6">
```
rsheeter, line 65: is "black" the shortest way to spell black? Is 3-digit hex legal, e.g. #000? ... maybe that's not enough difference to bother lol

PR #502, opened 2026-08-11
Migrate to mypy

rsheeter:
pytype evidently ends at 3.12 (https://github.com/google/pytype). https://github.com/fonttools/fonttools appears to use mypy, perhaps we should too?


rsheeter:
Closing in favor of pyrefly as apparently Google favors pyrefly and fonttools barely uses mypy so matching that is less important.

PR #508, opened 2026-09-13
Switch from deprecated "toml" library to tomllib and tomli-w

### LINKED ISSUE #451: Switch from deprecated `toml` library

hosiet:
The "toml" package is unmaintained and only implements TOML 0.5.0. Read config files with tomllib from the standard library (`tomli` on Python 3.10, same API) and write them with `tomli-w`.

Both return and accept plain Python types, so the tomlkit wrapper type problems that led to reverting #452 do not apply.

One behavioural detail: `toml.dumps()` silently skipped None values, while `tomli_w` raises TypeError on them. `config.write()` now drops unset optional values (clipbox_quantization by default) before serializing, so the written files are unchanged.

Tested with the full test suite as part of the Debian package (317 passed, 16 skipped).

fixes: #451

Downstream bug report: https://bugs.debian.org/1147582

Disclaimer: this pull request was drafted via Claude Fable 5.1 and then revised by human.

Please let me know your thoughts. And even better: if you could consider dropping the support to Python 3.10 (which will be EOL on Oct 2026), we can further drop the compatibility layer to `tomli` library and switch to `tomlib` + `tomli-w` completely.

cc: @anthrotype


a-detiste:
> Downstream bug report: https://bugs.debian.org/1147582

Hi, I'm impressed :-)

anthrotype:
thanks. I bumped `python_requires` separately in #509 and merged origin/main into yours. Will merge when CI goes green

